PR #26, opened 2022-06-16
USEID-125: Add widget.js to be embedded

code28:


src/main/resources/static/widget.js
```diff
@@ -0,0 +1,36 @@
+function getIframe() {
+  let tcTokenURL = new URL(document.currentScript.src).searchParams.get(
+    "tcTokenURL"
+  );
+  if (tcTokenURL === null) {
+    tcTokenURL = document.getElementById("useid-widget").dataset.tcTokenUrl;
+
+    if (tcTokenURL === undefined) {
+      const error = document.createElement("div");
+      error.innerHTML = "Fehlerhafte Konfiguration: TC Token nicht definiert.";
```
mpanne, line 10: The error should say "TC Token URL nicht definiert.", right?

src/main/resources/static/widget.js
```diff
@@ -0,0 +1,36 @@
...
+      return error;
+    }
+  }
+
+  const iframe = document.createElement("iframe");
+  iframe.setAttribute(
+    "src",
+    `https://useid.dev.ds4g.net/widget?tcTokenURL=${tcTokenURL}`
```
mpanne, line 18: The URL should somehow be injected depending on the environment instead of being hard coded? How is the script provided to the customers? Maybe the endpoint should be aware of the environment?

src/main/resources/static/widget.js
```diff
@@ -0,0 +1,36 @@
...
+  );
+  iframe.style.border = "2px solid #B8BDC3";
+  iframe.style.borderRadius = "20px";
+  iframe.style.width = "100%";
+  iframe.style.minHeight = "600px"; // TODO: Set height to fit
```
mpanne, line 23: I would leave styling to the surrounding container. Just use width and height 100%. Shouldn't that be enough?

src/main/resources/static/widget.js
```diff
@@ -0,0 +1,36 @@
...
+  return iframe;
+}
+
+const container = document.getElementById("useid-widget-container");
+container === null
+  ? document.write(getIframe().outerHTML)
```
mpanne, line 29: We could insert a default container here with some styling (see comment above)

src/main/resources/static/widget.js
```diff
@@ -0,0 +1,36 @@
...
+  : container.appendChild(getIframe());
+
+window.addEventListener("message", (e) => {
+  if (e.origin === "https://useid.dev.ds4g.net") {
```
mpanne, line 33: Same here with the URL (see comment above)

mpanne (changes requested):

src/main/resources/static/widget.js
code28: Right!

src/main/resources/static/widget.js
code28: Right now, it's only in the static folder. But of course you're right: We shouldn't hard code the URL here, so we probably need some template file and a route, injecting the env?

src/main/resources/static/widget.js
code28: Yes, agree. It was for now to have some nicer look until the design is done. Unfortunately, handling the height is not that easy, because the parent can't access the height of the iframe. So I think we should at least handle that, to show the full widget.

src/main/resources/static/widget.js
code28: So the parent page could still style it, even if they didn't implement it themselves? Yes, we can do that.

src/main/resources/static/widget.js
mpanne: Understand. Re height: Can we calculate the needed height according to the content of the iframe?

src/main/resources/static/widget.js
pgurusinga: or parse the current URL `window.location.href` and append the query param?

mpanne: The widget will be included in the eService page (window location would point to the eService domain). The URL must point to our backend though.

pgurusinga (approved):
perfect widget. love it that we use `??` 🎉 🎉

mpanne (approved):

PR #33, opened 2022-06-24
USEID-166: Add adr for URL format in QR code

code28:


pgurusinga (approved):

doc/adr/0006-url-format-in-qr-code.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Context
+
+The web widget displays a QR code, which transfers data to the mobile device, i.e. into the eID client.
+
+The QR code needs to include the ClientURL, including the `tcTokenURL`, (see [TR-03124 part 1](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03124/TR-03124-1.pdf) for more information).
+
+Additionally, some kind of widget session id needs to be added, so that the proxy message server can identify this exact widget when the mobile app wants to send a message to the widget.
+
+Lastly, an [encryption key](https://github.com/digitalservicebund/useid-architecture/blob/50b667dfe5d38846f3ac46ae8ab2a69c2ed63afa/doc/adr/0004-e2ee-for-refresh-address.md) needs to added. This should never be sent to any server and only be known to the widget and the mobile app.
```
code28, line 17: The link should be updated after [useid-architecture#5](https://github.com/digitalservicebund/useid-architecture/pull/5) was merged.

doc/adr/0006-url-format-in-qr-code.md
```diff
@@ -0,0 +1,49 @@
...
+
+The mobile flag (`{{isMobileDevice}}`) has boolean values (`true` or `false`).
+
+## Consequences
+
+The URL follows the Technical Guideline and is compatible to other eID clients.
+
+The URL contains all the needed data, including the `widgetSessionId` and the encryption key.
+
+The encryption key is not sent to any server, since it's added as fragment id.
```
mpanne, line 39: The ClientURL is never sent to any server. It is just encoded in the QR code and read by the mobile app. Why shouldn't the encryption key be included as a regular query parameter?

doc/adr/0005-csp-for-embedding-widget.md
```diff
@@ -0,0 +1,35 @@
...
+
+We use the [`frame-ancestors` directive at the CSP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors). The browsers only allow pages to embed an iframe, whose domain is listed there.
+
+We set this header only to the one domain, which is actually embedding the iframe and don't provide a full list of all allowed domains.
+
+The page inside the iframe can't get the domain of the parent page. Therefore the [widget.js file](0004-embedding-widget-via-js-wrapper.md) adds the domain of the parent page as query parameter to the iframe url.
+
+The backend can check the domain from the query parameter with the list of registered domains and if the domain is found, it is added into the CSP header.
+
+We evaluated using a kind of api key (which would then be public), but creating a mapping of domain to api key would be quite easy, so we use the domain name as quasi public api key.
```
mpanne, line 27: Using an API key instead of the domain would make it harder (or nearly impossible) to use this method to observe which domains are registered with UseID. I do not know if that's relevant though (could be relevant for attackers who want to find a vulnerable domain or something like that).

doc/adr/0005-csp-for-embedding-widget.md
```diff
@@ -0,0 +1,35 @@
+# 5. CSP for embedding widget
+
+Date: 2022-06-24
+
+## Status
+
+Proposed
```
mpanne, line 7: Let's decide and set this to "Accepted"?

mpanne (changes requested):
Let's clarify open points and set both ADRs to "Accepted" before merging I would propose.

doc/adr/0006-url-format-in-qr-code.md
code28: Good question. This whole URL makes actually no sense at all. But since it now is an URL, we can be double sure for any future changes by using fragment id. I'll add that.

hannesschaletzky (approved):

pgurusinga (approved):

PR #34, opened 2022-06-28
USEID-131: Implement SSE to specific client

mpanne:


src/main/kotlin/de/bund/digitalservice/useid/events/Event.kt
```diff
@@ -0,0 +1,23 @@
+package de.bund.digitalservice.useid.events
+
+class Event(val consumerId: String, val encryptedRefreshAddress: String, val widgetSessionId: String) {
```
code28, line 3: I thought that `consumerId` and `widgetSessionId` are the same. Maybe we should talk about, who creates which id and when.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -1,12 +1,12 @@
...
+package de.bund.digitalservice.useid.widget
 
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
 import org.springframework.web.bind.annotation.GetMapping
 import reactor.core.publisher.Mono
 
 @Controller
-class QRCodeController {
+class WidgetController {
```
code28, line 9: Not quite in this scope, but anyway: 👍

src/main/kotlin/de/bund/digitalservice/useid/events/EventType.kt
```diff
@@ -0,0 +1,5 @@
+package de.bund.digitalservice.useid.events
+
+enum class EventType(val eventName: String) {
+    SUCCESS("success")
```
code28, line 4: Do we still need this?

src/main/kotlin/de/bund/digitalservice/useid/events/EventHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+        log.info { "New consumer added with id $id, total consumers: ${consumers.size}" }
+    }
+
+    /**
+     * This method publishes the given event to the according consumer. The consumer id is specified in the event.
+     */
+    fun publish(event: Event) {
+        val consumerId = event.consumerId
+        log.info { "Publish event to consumer $consumerId" }
+        consumers[consumerId]!!.accept(event)
```
code28, line 28: Maybe we should make sure, that the consumerId is available in that map? (In a more.. graceful way?)

code28 (changes requested):
Looks good to me! Great that you made it work at all.
A few small comments - we should probably talk and define the payload.

src/main/kotlin/de/bund/digitalservice/useid/events/Event.kt
pgurusinga: @code28 I thought so too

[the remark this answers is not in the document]
mpanne: They are only used by the tests actually. I didn't know about `data class` before, it's a Kotlin feature. Great hint, I will change it!

src/main/kotlin/de/bund/digitalservice/useid/events/EventType.kt
mpanne: I am not sure. Currently we only talked about the happy path of the whole flow. What if errors occur in the eID-Client when reading the data? Do we inform the eService about it? Or do we only send a SSE in case of success?

mpanne: Currently this type is used as event type for the SSE. The enum only makes sense in case we add other types later on.

src/main/kotlin/de/bund/digitalservice/useid/events/Event.kt
mpanne: Good point. I will consolidate those two.

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * This endpoint receives events from the eID client (i.e. Ident-App) and publishes them to the respective consumer.
+     */
+    @PostMapping("/events")
+    @ResponseStatus(HttpStatus.CREATED)
+    fun send(@RequestBody event: Event): Mono<Event> {
+        log.info { "Received event for consumer: ${event.widgetSessionId}" }
+        eventHandler.publish(event)
+        return Mono.just(event)
```
pgurusinga, line 38: As far as I know, client (browser) won't consume any response, right? I think we can just return`Mono.empty()` instead? what do you think?

src/main/kotlin/de/bund/digitalservice/useid/events/EventHandler.kt
code28: When the `consumerId` is not available in the map `consumer`, we throw a `NullPointerException` (`!!`), right? I suggested, that we could handle this event maybe more graceful and don't return a general NPE, i.e.

* either handle the exception and return some kind of application-error (`WidgetNotFound` or something), so that the app can show some other error message than if a NPE exception happens at some other code meaning something different
* or prevent that exception by just ignoring the missing value inside the map (so that someone brute-forcing can't get any information at all, but our app won't learn anything either)

src/main/kotlin/de/bund/digitalservice/useid/events/EventType.kt
code28: Good point. We should inform the widget also in error cases, so that we can display some error message.

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
code28: This endpoint is called by the mobile app. It should consume the response, so that we can return errors if some happen, but in the happy case, we don't need to return anything probably.

mpanne: True, I will return empty for now.

src/main/kotlin/de/bund/digitalservice/useid/events/EventType.kt
mpanne: I see this as an extension and not necessarily in the scope of this PR. I would error cases when we need them and leave it like that for now. What do you think?

carhartl (approved):
With `build.gradle.kts` being changed the checksum in `.talismanrc` will have to be updated - the secrets check currently only runs against the main branch: https://github.com/digitalservicebund/useid-backend-service/blob/8fa08647d0e4eeb948f5322e7eb35956af1a7975/.github/workflows/secrets-check.yml#L4-L5

Here's a snippet that I use to populate the clipboard with the required checksum:
```bash
talisman -c build.gradle.kts | sed -n '3,5p' | yq e '.fileignoreconfig[0].checksum' - | tr -d '\r\n' | pbcopy
```

(You'll need `brew install talisman yq`)

With the code from the prototype removed we can clean up a bit of build configuration (that address skewed coverage metrics) and I amended the PR with a commit for this. I've update the `.talismanrc` along with it.

On a general note: a PR with 20 commits and no description can be a bit frightening :)

src/main/kotlin/de/bund/digitalservice/useid/events/EventType.kt
code28: Yes, I agree. Didn't want to say anything different, just keeping the thought at the back of my head to handle error messages later :)

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
carhartl: Yes, semantically that seems more appropriate for a 201. Thinking about this, a 202 ACCEPTED might be more correct technically. We can only say for sure that we retrieved the request and deal with it.

pgurusinga: > This endpoint is called by the mobile app. It should consume the response, so that we can return errors if some happen, but in the happy case, we don't need to return anything probably.

@code28 you are right! I think we could also return nothing as payload in case of error and a proper status code (4xx, 5xx), then the client can decide how to handle the status code, e.g. the client can load the error message from translation.. but this idea is probably not for now but future improvement.

src/main/kotlin/de/bund/digitalservice/useid/events/EventHandler.kt
mpanne: Good point! I like the idea of ignoring unknown IDs since it keeps it very simple and prevents observing IDs via brute force. If we assume that most of the messages sent to this endpoint are from a valid source (not an attacker) though, we must assume that the messages are required to complete an ongoing flow and ignoring them would mean missing the success information since neither the eID client nor our server will notice (i.e. no retry or error handling possible) which is why I would go for a more descriptive error response.

code28: Exactly my train of thought. 🤓 With this, maybe we can handle users closing the browser window or something by displaying a helpful message in the app?!

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
mpanne: The endpoint returns status 202 ACCEPTED with an empty Mono now.

build.gradle.kts
```diff
@@ -49,13 +49,18 @@ dependencies {
...
     implementation("com.google.zxing:javase:3.5.0")
     implementation("org.springdoc:springdoc-openapi-webflux-ui:1.6.9")
     runtimeOnly("org.springdoc:springdoc-openapi-kotlin:1.6.9")
 
     developmentOnly("org.springframework.boot:spring-boot-devtools")
 
-    testImplementation("org.springframework.boot:spring-boot-starter-test")
+    testImplementation("org.springframework.boot:spring-boot-starter-test") {
+        exclude("org.mockito", "mockito-core")
+        because("Use MockK instead of Mockito since it is better suited for Kotlin")
```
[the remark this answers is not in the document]
mpanne: Done: https://github.com/digitalservicebund/kotlin-application-template/commit/39eaf8b27dc8e00c784b797931781c5fc5d04a53

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
```diff
@@ -0,0 +1,66 @@
...
+    @PostMapping("/events")
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    fun send(@RequestBody event: Event): Mono<ResponseEntity<ErrorResponseBody>> {
+        log.info { "Received event for consumer: ${event.widgetSessionId}" }
+        try {
+            eventHandler.publish(event)
+        } catch (e: ConsumerNotFoundException) {
+            return createNotFoundResponse(e.message!!)
+        }
+        return Mono.empty()
```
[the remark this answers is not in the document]
carhartl: Try:

```java
Mono.fromCallable(() -> eventHandler.publish(event))
  .map((next) -> /* successful response */)
  .onErrorReturn(ex -> Mono.just(createNotFoundResponse(e.message!!)))
```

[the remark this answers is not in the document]
carhartl: (untested)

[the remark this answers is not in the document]
carhartl: I looked into this using Kotlin, here's how you could do it:

```diff
diff --git a/src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt b/src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
index 5835176..467a7aa 100644
--- a/src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
@@ -33,14 +33,16 @@ class EventController(eventHandler: EventHandler) {
      */
     @PostMapping("/events")
     @ResponseStatus(HttpStatus.ACCEPTED)
-    fun send(@RequestBody event: Event): Mono<ResponseEntity<ErrorResponseBody>> {
+    fun send(@RequestBody event: Event): Mono<ResponseEntity<Any>> {
         log.info { "Received event for consumer: ${event.widgetSessionId}" }
-        try {
-            eventHandler.publish(event)
-        } catch (e: ConsumerNotFoundException) {
-            return createNotFoundResponse(e.message!!)
-        }
-        return Mono.empty()
+        Mono.fromCallable { eventHandler.publish(event) }
+            .map { ResponseEntity.accepted() }
+            .onErrorResume {
+                Mono.just(
+                    ResponseEntity
+                        .status(HttpStatus.NOT_FOUND)
+                )
+            }
     }
 
     /**
@@ -57,10 +59,4 @@ class EventController(eventHandler: EventHandler) {
         .data(event)
         .event(EventType.SUCCESS.eventName)
         .build()
-
-    private fun createNotFoundResponse(message: String) = Mono.just(
-        ResponseEntity
-            .status(HttpStatus.NOT_FOUND)
-            .body(ErrorResponseBody(HttpStatus.NOT_FOUND.value(), message))
-    )
 }
```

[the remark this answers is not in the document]
carhartl: >Is it possible to write it like this? I assume try... catch... in Reactor would do blocking operation...

It's only blocking if `eventHandler.publish(event)` was doing a blocking operation, such as IO, network call etc.

But the code at hand -- try/catch in general -- isn't very idiomatic in terms of reactiveness and when using it alongside some reactive operations more often it's a smell at least.

pgurusinga (approved):

PR #36, opened 2022-06-30
USEID-157 QR Code Generation in Front-End instead of Back-End

pgurusinga:


doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+
+Date: 2022-06-30
+
+## Status
+
+Accepted
+
+## Context
+
+QR Code is embedded in the eService page and it is used to allow user to start identification process via the UseID app camera which later. The QR Code contains `tcTokenURL`, `widgetSessionId` and `encryptionKey`. For this decision record, [ADR 0006](LINK TO 0006)
```
code28, line 11: There are probably a few words too much? "camera which later"
But it should better be "camera app" instead of "UseID app".
And the link intro should be: "For this decision, see record ..."

So all in all maybe:
```suggestion
A QR code is embedded in the eService page via iframe widget and it is used to allow user to start identification process via camera app on a mobile device. The QR Code contains `tcTokenURL`, `widgetSessionId`, a `mobile` flag and an `encryptionKey`. (For details see [ADR 0006](LINK TO 0006).)
```

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+QR Code is embedded in the eService page and it is used to allow user to start identification process via the UseID app camera which later. The QR Code contains `tcTokenURL`, `widgetSessionId` and `encryptionKey`. For this decision record, [ADR 0006](LINK TO 0006) 
+```
+eid://127.0.0.1:24727/eID-Client?tcTokenURL={{tcTokenURL}}&widgetSessionId={{widgetSessionId}}&mobile={{isMobileDevice}}#{{encryptionKey}}
```
code28, line 13: Not sure, if we should add the URL format here again? Makes it harder to change it, because more than 1 adr needs to be changed then. I would just remove it, since here it's not about the content of the QR code and only about where to create it. What's inside the QR code is already listed above.

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+```
+The current QR Code generation is handled in the server `/v1/qrcode/{imageSize}`
```
code28, line 15: Since we have no ADR about that, I'd just not mention it. We still should add some context, why there is a problem:
> The `encryptionKey`, which needs to be encoded in the QR code, must not leave the user's device, i.e. it must not be sent to a server.

And maybe add something about the `tcTokenURL`, that it contains information about the session, the server should (or must) not be aware of?

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+
+## Decision
+
+In order to improve the security measures, QR Code should be handled on the client side.
```
code28, line 19: I think, we could formulate it like proposed with a "we" sentence in the decision section, so e.g.
```suggestion
We generate the QR code on the client side. This improves the security, since we don't need to send any data to the server.
```

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+1. eService provides the necessary information  `tcTokenURL`, `widgetSessionId` and `encryptionKey`
```
code28, line 20: That's not correct. The eService only provides the `tcTokenURL`. The `widgetSessionId` is created by the UseID backend and injected into the widget, the `encryptionKey` is generated client-side inside the widget.

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+2. Widget provides the eService with QR Code library
```
code28, line 21: That's not correct. The widget uses a QR code library to generate the QR code and just displays it. There is no link to the eService to send the QR code.

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+
+## Consequences
```
code28, line 23: There should be an empty line after a headline.

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+1. Increase load time
```
code28, line 24: Maybe add a reason?

> We need to load a QR code generation library and generate it, therefore the page load time increases.

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
...
+2. UseID Backend will not have any knowledge of `tcTokenUrl`
```
code28, line 25: "...and `encryptionKey`." ?

code28 (changes requested):

doc/adr/0007-generate-qr-code-on-client
pgurusinga: I think if we would like to change the scheme, we should create a new ADR entry in the future

doc/adr/0007-generate-qr-code-on-client
```diff
@@ -0,0 +1,25 @@
+# 1. Record architecture decisions
```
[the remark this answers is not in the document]
pgurusinga: 🙈 🙈

mpanne (changes requested):
Thanks for the ADR! 
As mentioned in the comments, we (Hannes and me) would suggest to have an eye on proper english language to ensure a good readability. 
Cheers!

mpanne (approved):

hannesschaletzky (approved):

PR #41, opened 2022-07-06
USEID-158 Add ADR for API keys

pgurusinga:


doc/adr/0009-api-key.md
```diff
@@ -0,0 +1,26 @@
...
+
+Date: 2022-07-06
+
+## Status
+
+Accepted
+
+## Context
+
+We want to filter incoming requests in the UseID backend that only client with an approved integration could access UseID endpoints.
```
mpanne, line 11:
```suggestion
We want to filter incoming requests in the UseID backend, so that only authorized clients are able to access the API.
```

doc/adr/0009-api-key.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Decision
+
+We use Bearer authentication. The Client (eService) should add a token in the Authorization header when requesting to the endpoints and the token should be a random string of 51 characters.
```
mpanne, line 15:
```suggestion
We use Bearer authentication. The Client (eService) should add a token in the Authorization header when requesting to the endpoints and the token should be a string of 51 characters which was issued to the client by us.
```

doc/adr/0009-api-key.md
```diff
@@ -0,0 +1,26 @@
...
+
+e.g.
+
+`Authorization: Bearer <token>`
+
+The token should, for MVP purpose, be configured as environment variable. For the future releases, a lifecycle of token should be defined, e.g. validity of the token.
```
mpanne, line 21: I would remove the whole sentence since it is an implementation detail which is not relevant for the decision.

doc/adr/0009-api-key.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Consequences
+
+1. Reduce our security risk, since we only allow known client in the UseID backend
```
mpanne, line 25:
```suggestion
1. Reduce our security risk, since we only allow authorized clients to access the API
```

doc/adr/0009-api-key.md
```diff
@@ -0,0 +1,26 @@
...
+2. Reduce resource usage, since we are filtering unknown requests, and we do not process them further
```
mpanne, line 26: Not relevant. Would remove. I think the API key is for security reasons only.

mpanne (changes requested):

mpanne (approved):

hannesschaletzky (approved):

PR #45, opened 2022-07-25
USEID-249: Add Basic Auth for API

pgurusinga:
Basic Authentication for the Identification sessions and Useid Infra is set with the required env vars

### DIFF
--- a/src/test/resources/application.properties
+++ b/src/test/resources/application.properties
@@ -1,3 +1,3 @@
 application.staging.url=${STAGING_URL:http://localhost:8080}
-api.user.username=${ACTIONS_TEST_USERNAME}
-api.user.password=${ACTIONS_TEST_PASSWORD}
+api.user.username=${ACTIONS_TEST_USERNAME:testUser}
+api.user.password=${ACTIONS_TEST_PASSWORD:testPassword}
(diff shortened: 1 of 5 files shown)


src/test/resources/application.properties
```diff
@@ -1 +1,3 @@
 application.staging.url=${STAGING_URL:http://localhost:8080}
+api.user.username=${api_user_username}
```
code28, line 2: I assume, those are environment variables? We should keep an consistent naming scheme, so UPPERCASE please, since `STAGING_URL` is already upper case.

code28 (changes requested):

src/test/resources/application.properties
```diff
@@ -1 +1,3 @@
 application.staging.url=${STAGING_URL:http://localhost:8080}
+api.user.username=${ACTIONS_TEST_USERNAME}
```
code28, line 2: Why did you call the env variable like that and not just `API_USER_USERNAME`?

code28 (changes requested):

src/test/resources/application.properties
pgurusinga: it was also my thought, but then this env var is required during build, so I name it with easier to think about `ACTIONS_TEST_USERNAME` means that this env will be available only in Github Actions and the setting is available in Github, but I can rename it as well to reflect what is currently set in our infra

code28: I see your point. Maybe then just add a default value, since this is only used for tests?

hannesschaletzky (approved):

PR #50, opened 2022-08-05
USEID-216: Secure identification session API with API Keys

mpanne:
**Goal**
When GST is using the APIs, we want to have an authentication via API keys in place so that we can ensure a secure and exclusive API access for GST. 

**Implementation**
* Create an “ApiKeyAuthenticationToken” which implements the “Authentication” interface and represents the current authentication of the entity accessing the API with the API key
* Create “ServerAuthenticationConverter” to extract the API token from the request header and create the unauthenticated “ApiKeyAuthenticationToken”
* Create “ReactiveAuthenticationManager” to validate the API key and authenticate the “Authentication” object which was created from the converter
* Create “AuthenticationWebFilter” with “ServerAuthenticationConverter” and “ReactiveAuthenticationManager” which filters all requests accessing the identification session API


src/test/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationConverterTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        // When
+        val authenticationMono = authenticationConverter.convert(serverWebExchange)
+
+        // Then
+        StepVerifier.create(authenticationMono)
+            .expectNextCount(0)
+            .verifyComplete()
+    }
+
+    private fun invalidHeaders(): Stream<Arguments?>? {
```
code28, line 65: Nice shortcut!

src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsControllerIntegrationTest.kt
```diff
@@ -41,7 +38,7 @@ class IdentificationSessionsControllerIntegrationTest(
         webTestClient
             .post()
             .uri(URI.create("http://localhost:$port/api/v1/identification/sessions"))
-            .headers { headers -> headers.setBasicAuth(username, password) }
+            .headers { headers -> setAuthorizationHeader(headers) }
```
code28, line 41: Maybe we can have this shorter? I think the method name makes it clear, that the value are headers, so we could use `.headers { setAuthorizationHeader(it) }`? But I'm fine with the a bit more verbose solution, too.

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
```diff
@@ -0,0 +1,35 @@
+package de.bund.digitalservice.useid.apikeys
+
+import org.springframework.security.core.Authentication
+import org.springframework.security.core.GrantedAuthority
+
+class ApiKeyAuthenticationToken(private val apiKey: String, private var authenticated: Boolean = false) : Authentication {
```
code28, line 6: Didn't you want to add a URL (static refresh address) to this token? Or is that for another PR?

code28 (changes requested):
Just tiny comments - can be resolved with comments or code :)

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
code28: Now the refresh address is in the apiKeys-List, but here still missing? Or is it accessed differently?

src/main/resources/application.yaml
```diff
@@ -0,0 +1,19 @@
+management:
```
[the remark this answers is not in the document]
pgurusinga: ah yes, somehow the configuration file cannot be loaded in test environment even though there was a way, but as I understand, we need to write code configuration (no first-class support) and I decided to not use the YAML and keep the `.properties` instead.
https://stackoverflow.com/questions/21271468/spring-propertysource-using-yaml

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationConverter.kt
```diff
@@ -0,0 +1,35 @@
...
+            .flatMap { serverWebExchange: ServerWebExchange ->
+                Mono.justOrEmpty(serverWebExchange.request.headers.getFirst(HttpHeaders.AUTHORIZATION))
+            }
+            .filter { header: String ->
+                header.trim().isNotEmpty() && header.trim().startsWith(AUTH_HEADER_VALUE_PREFIX)
+            }
+            .map { header: String ->
+                header.substring(AUTH_HEADER_VALUE_PREFIX.length)
+            }
+            .switchIfEmpty(Mono.empty())
```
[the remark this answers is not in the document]
mpanne: You are probably right. I will test it and remove it in case.

src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsControllerIntegrationTest.kt
mpanne: Good catch. Done.

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationConverter.kt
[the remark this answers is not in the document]
mpanne: Done.

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
mpanne: I'm uncertain. Either we store it there (and return it via `getDetails`) or we look it up in the `apiProperties` when it's needed (i.e. when the `tcToken` is created). What do you think?

src/main/resources/application.yaml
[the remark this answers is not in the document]
code28: 👍

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
code28: Hmm.. When looking it up in the `apiProperties` later, we'd have to use the `apiKey` as key, right? I think, I'd rather have it here and getting it via `getDetails`. Then we only have one class, which fetches values from `apiProperties` and the details about the authenticated "user" (which is for now only this URL) are all in this token.

mpanne: I will put it here. 
This class is not the only one fetching data from the `apiProperties` though. The `ApiKeyAuthenticationManager` does that too. But storing all "user" information in one place convinces me.

src/main/resources/application.yaml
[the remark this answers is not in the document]
mpanne: If there are no objections agains YAML, I would leave it like that.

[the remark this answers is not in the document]
code28: Yes, please leave YAML.

.talismanrc
```diff
@@ -13,7 +13,20 @@ fileignoreconfig:
...
+  - filename: src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationManager.kt
+    checksum: c784f489b09de0e3de46ca9e8f6d52521f557542b3b37c82032b13c3c9c4cbff
+  - filename: src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationConverter.kt
+    checksum: 2881fab3f14e061c4a3327046e5376a1b5f1841cd57c67edc6aa05dd07f84b82
+  - filename: src/test/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationManagerTest.kt
+    checksum: d5b224b9b4f8f32193dc8d997e5b2a2c24df79cbbc115ff73701c5676ff7cf63
+  - filename: src/test/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationConverterTest.kt
+    checksum: 7acd129e6ce59738f1b819adc2b482b49a43ffdbe1e21a33bbb33443afaed88d
+  - filename: src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
+    checksum: 61c0baa54d9ef1e1281e19db74dcb8be96c37be2b09c41e4d8d9bdca8865f69c
```
pgurusinga, line 31: is it possible to use glob in here? it's similar to the fonts above with asterisk `"*.woff*"`
so could we try to use `src/main/kotlin/de/bund/digitalservice/useid/apikeys/*` instead?

src/main/kotlin/de/bund/digitalservice/useid/apikeys/ApiKeyAuthenticationToken.kt
mpanne: Done, please check again.

.talismanrc
mpanne: Good hint working with wildcards. Makes the file much better readable.

code28 (approved):
Yes! Nice!

pgurusinga (approved):

PR #52, opened 2022-08-11
USEID-230 Implement tcToken endpoint

pgurusinga:
https://digitalservicebund.atlassian.net/jira/software/projects/USEID/boards/3?selectedIssue=USEID-230

### Proposed changes:
1. `tcTokenUrl` is returned to the eID-client
2. eID-client is able to receive tcToken information from our service

### Exclusion:
We decided to finish this ticket earlier and leave out two things: 
1. Mock Integration Tests to  the eIDServer in IdentificationSessionsControllerTest
2. Provide environment variables through pipeline (currently default values in application.yaml)


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -1,26 +1,33 @@
 package de.bund.digitalservice.useid.identification
 
+import de.bos_bremen.gov.autent.common.Utils
```
code28, line 3: Huh, weird import name 😅

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -1,26 +1,33 @@
...
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.util.UriComponentsBuilder
 import reactor.core.publisher.Mono
+import reactor.core.scheduler.Schedulers
+import java.net.URLEncoder
 import java.util.UUID
 
 internal const val IDENTIFICATION_SESSIONS_ENDPOINT = "/api/v1/identification/sessions"
+internal const val TCTOKEN_ENDPOINT = "tc-token"
```
code28, line 24: Maybe we should rename the above as "prefix"? Otherwise, this looks a bit weird here and you could think that the endpoints are `/api/v1/identification/sessions` and `/tc-token`, even though the latter is `/api/v1/identification/sessions/tc-token".

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -35,40 +42,67 @@ class IdentificationSessionsController(
...
+        serverHttpRequest: ServerHttpRequest
+    ): Mono<ResponseEntity<CreateIdentitySessionResponse>> {
+        return identificationSessionService
+            .save(
+                refreshAddress = "https://localhost:8443/", // Currently a mock value, later will inject from env var mapped by API Key
+                requestAttributes = createIdentitySessionRequest.requestAttributes
+            ) // PublishOn is suggested by IntelliJ, maybe there is another way to resolve blocking URLEncoder.encode()
+            .publishOn(Schedulers.boundedElastic())
+            .map {
+                val encodedTcTokenUrl = URLEncoder.encode("${serverHttpRequest.uri}/$TCTOKEN_ENDPOINT/${it.useIDSessionId}", Utils.ENCODING)
```
code28, line 56: Is this URL working? From the code, I'd assume, that it isn't. It's missing `/api/v1/identification/sessions`.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -35,40 +42,67 @@ class IdentificationSessionsController(
...
+    }
+
+    @GetMapping("/$TCTOKEN_ENDPOINT/{useIDSessionId}")
+    fun getTCToken(@PathVariable useIDSessionId: UUID): Mono<ResponseEntity<TCTokenType>> {
+        return identificationSessionService.findByIdOrFail(useIDSessionId)
+            .flatMap {
+                tcTokenService.getTcToken(it.refreshAddress)
+            }
+            .doOnNext {
+                val eIDSessionId = UriComponentsBuilder
```
code28, line 79: Phew... 😁

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
```diff
@@ -1,23 +1,45 @@
...
+
+    fun findByIdOrFail(useIDSessionId: UUID): Mono<IdentificationSession> {
+        if (!sessionExists(useIDSessionId)) {
+            throw Error("no session found with useIDSessionId $useIDSessionId")
+        }
+        return Mono.just(sessions.find { it.useIDSessionId == useIDSessionId }!!)
+    }
+
+    fun updateEIDSessionId(useIDSessionId: UUID, eIDSessionId: UUID) {
+        // use !! here bc findByIdOrFail was already executed
```
code28, line 38: We shouldn't do such a thing, because this method could be used in the future somewhere else or in  a different way and this wouldn't hold. So either pass in the session, which was received by `findByIdOrFail` or call `findByIdOrFail` again.

src/main/kotlin/de/bund/digitalservice/useid/identification/TcTokenService.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val log = KotlinLogging.logger {}
+    fun getTcToken(refreshAddress: String): Mono<TCTokenType> {
+        /* TODO: current implementation is a blocking operation
+        * We cannot just wrap a blocking call with a Mono or a Flux
+        * https://betterprogramming.pub/how-to-avoid-blocking-in-reactive-java-757ec7024676
+        * and also usage of Reactor's BlockHound is preferred
+        */
+        return Mono.just(eidService.getTcToken(refreshAddress))
+            .doOnNext {
+                log.info("created tcToken ${XmlHelper.marshalObject(it)}")
```
code28, line 23: I don't think we need a log here, especially not "info", since this route will be one of the most called ones.

src/main/kotlin/de/bund/digitalservice/useid/utils/XmlHelper.kt
```diff
@@ -0,0 +1,30 @@
+package de.bund.digitalservice.useid.utils
+
+import java.io.StringWriter
+import javax.xml.bind.JAXBContext
+import javax.xml.bind.JAXBException
+import javax.xml.bind.Marshaller
+
+object XmlHelper {
```
code28, line 8: See above comment. Then we can remove this.

code28 (changes requested):

allowed-licenses.json
```diff
@@ -3,6 +3,9 @@
     {
       "moduleLicense": "Apache License, Version 2.0"
     },
+    {
+      "moduleLicense": "Bouncy Castle Licence"
```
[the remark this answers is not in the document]
pgurusinga: there was some packages that requires this license, but it's not related to SDK. the license is same as MIT https://www.bouncycastle.org/licence.html. So I just removed the Bouncy Castle license locally again, it does not throw error, it probably was a part of SDK dependencies.

update: I remove the SDK license (So no changes in this file anymore) and I ran check license task:
image
I am so confused now. 
cc: @hannesschaletzky

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
pgurusinga: yes, this URL is working, my assumption since client requests to the endpoint `/api/v1/identification/sessions`, the `serverHttpRequest.uri` contains the correct path already.

image

src/main/kotlin/de/bund/digitalservice/useid/identification/TcTokenService.kt
pgurusinga: probably not for `success` case and I agree the XML helper is for debugging purpose, but in case `getTcToken` returns error (e.g server error), i think we need to log it for e.g Sentry? what do you think?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
code28: Wow, that's.. unexpected. And a bit error-prone, since we might change the route of this endpoint, so that it's no prefix of the `tcTokenUrl` anymore? Hmm... not sure about it here, but this feels too "magic" for me for being self-explanatory.

src/main/kotlin/de/bund/digitalservice/useid/identification/TcTokenService.kt
code28: Yes, good point, we need to log the response of the eIDServer in an error case (but still probably not `info`, maybe even higher?).

pgurusinga: I created a story for Sentry, because it will increase this PR and we should properly discuss what error we want to observe => https://digitalservicebund.atlassian.net/browse/USEID-312

I would remove this `log` and the `XMLHelper` for now.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -35,40 +42,67 @@ class IdentificationSessionsController(
...
+                log.error {
+                    "error occurred when creating identification session: ${exception.message}"
+                }
+            }
+            .onErrorReturn(
+                ResponseEntity.internalServerError().body(null)
+            )
+    }
+
+    @GetMapping("/$TCTOKEN_ENDPOINT/{useIDSessionId}")
```
mpanne, line 72: That's not the REST way to do it. It's always `/resource/{resource-id}/resource` so it should be `/sessions/{sessionId}/tc-token`

mpanne: Ah and please add the `produces` parameter of the `GetMapping` annotation to indicate that XML is returned.

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
```diff
@@ -1,23 +1,45 @@
...
 import java.util.UUID
 
 @Repository
 class MockDatasource {
     private val sessions = mutableListOf<IdentificationSession>()
+    private val log = KotlinLogging.logger {}
 
-    fun save(tcTokenUrl: String, refreshAddress: String, requestAttributes: List<String>): Mono<IdentificationSession> {
-        val currentSessionId = UUID.randomUUID()
+    fun save(refreshAddress: String, requestAttributes: List<String>): Mono<IdentificationSession> {
```
mpanne, line 13: Rename `save` to `create` since "save" indicates that you could also do updates with it (at least that's how Spring uses "save") but here a new session is always added.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
pgurusinga: I agree that the code is not really straight forward and we have to guess whether the `uri` contains the path or not, we can make it more explicit though.. since I am lack knowledge of Spring API, we will find a way

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
mpanne: +1 strongly agree! A method should never depend on another method to be executed before.

mpanne: And it does not hurt to call `findByIdOrFail` again here.

src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsControllerIntegrationTest.kt
```diff
@@ -24,7 +23,7 @@ class IdentificationSessionsControllerIntegrationTest(
     val attributes = listOf("DG1", "DG2")
 
     @Test
-    fun `requesting identification sessions returns 401 when the request is made without authentication`() {
+    fun `requesting identification sessions returns 401 when the request is made without basic authentication`() {
```
mpanne, line 26: Cleanup. ;) Probably a result of main merged in.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
mpanne: Yes, let's make it explicit please. I guess this would also break if we hide our service behind a proxy (or something like that) and would return internal domain names.

mpanne (changes requested):

.talismanrc
```diff
@@ -1,10 +1,12 @@
 fileignoreconfig:
   - filename: "*README.md"
     ignore_detectors: [filecontent]
+  - filename: src/main/kotlin/de/bund/digitalservice/useid/eidservice/EidWebService.kt
```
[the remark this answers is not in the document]
pgurusinga: thanks for the hint, this checksum is invalid because we rename the file 😓

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -35,40 +42,67 @@ class IdentificationSessionsController(
...
-                    createIdentitySessionRequest.requestAttributes
-                )
-            }.map {
+    fun createSession(
+        @RequestBody createIdentitySessionRequest: CreateIdentitySessionRequest,
+        serverHttpRequest: ServerHttpRequest
+    ): Mono<ResponseEntity<CreateIdentitySessionResponse>> {
+        return identificationSessionService
+            .save(
+                refreshAddress = "https://localhost:8443/", // Currently a mock value, later will inject from env var mapped by API Key
```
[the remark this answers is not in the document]
pgurusinga: ah is this something that will be available from the API story you did?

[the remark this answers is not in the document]
mpanne: Exactly :) it is already available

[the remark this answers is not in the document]
mpanne: Should work like this but you need to test. 

Screenshot 2022-08-12 at 15 57 40

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSession.kt
```diff
@@ -3,8 +3,8 @@ package de.bund.digitalservice.useid.identification
 import java.util.UUID
 
 data class IdentificationSession(
+    var eIDSessionId: UUID = UUID(0, 0),
```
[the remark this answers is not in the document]
pgurusinga: this assignment is meant for setting up initial value for the eIDSessionId, because `eIDSessionId` needs an initial value when we want to init the `IdentificationSession`, so that later in the future when we successfully fetched the `eIDSessionId` (from the tcToken endpont) we update its value. If we don't set an initial value, it throws an error in the `mockDataSource`. or is there a way to do that without assigning initial value?

src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsControllerIntegrationTest.kt
pgurusinga: yes again, the huge mess from the merge conflict 😓

build.gradle.kts
```diff
@@ -84,6 +90,9 @@ tasks {
         useJUnitPlatform {
             excludeTags("integration", "journey")
         }
+        testLogging {
+            events("passed", "skipped", "failed")
```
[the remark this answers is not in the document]
mpanne: Let's talk about trunk based development again soon...

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
pgurusinga:
```kotlin
internal const val PREFIX_IDENTIFICATION_ENDPOINT = "/api/v1/identification"
internal const val SESSIONS_ENDPOINT = "$PREFIX_IDENTIFICATION/sessions"
internal const val TCTOKEN_ENDPOINT = "tc-token"
```
or 
```kotlin
internal const val PREFIX= "/api/v1/identification/sessions"
internal const val TCTOKEN_ENDPOINT = "tc-token"
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
pgurusinga: @mpanne is this the right way to get query param in Spring? or is there another way?

code28: That part looks fine for me, but it's a shame, that we can't get that ID from the token directly. I don't understand at the moment, why the sessionIdentifier is a different one, maybe we can find out it about it (later). It's just a weird way to get that id, even though I don't see a better one at the moment without researching about that behavior. Maybe you can write a research story for this, so that we can have a look at a later stage? Because it's working for now.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSession.kt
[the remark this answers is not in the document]
code28: That's not good having magic values. Kotlin has something called nullable types.
```suggestion
    var eIDSessionId: UUID?
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
pgurusinga: @code28 also the expiration of the sessionId , which we don't know as well.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSession.kt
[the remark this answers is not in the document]
pgurusinga: yes, I had a look with the nullable type, but in the `MockDataSource` it throws an error
image

which then I need to assign default value again in the `MockDataSource`, if not, the `MockDataSource` will throw an error same as above.

```kotlin
val identificationSession = IdentificationSession(
    eIDSessionId = UUID(0,0),
    refreshAddress = refreshAddress,
    requestAttributes = requestAttributes,
    useIDSessionId = useIDSessionId
)
```
Why UUID(0, 0) ?
https://stackoverflow.com/questions/26637225/how-can-i-create-an-empty-java-util-uuid-object

or maybe I am missing something? 🤔

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
mpanne: I would suggest: 
```
internal const val IDENTIFICATION_SESSIONS_BASE_PATH = "/api/v1/identification/sessions"
internal const val TCTOKEN_PATH_SUFFIX = "tc-token"
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
mpanne: I agree with Simon that we should find out more about the sessionIDs. But the way you extract the query parameter is fine.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSession.kt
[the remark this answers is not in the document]
mpanne: That should do the trick. The value should be null until it is used. Otherwise it's hard to tell if the ID is already set or not. 

```suggestion
    var eIDSessionId: UUID? = nul l,
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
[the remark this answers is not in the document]
pgurusinga: yes, it works! 🎉

src/main/kotlin/de/bund/digitalservice/useid/identification/CreateIdentitySessionRequest.kt
```diff
@@ -1,6 +1,5 @@
 package de.bund.digitalservice.useid.identification
 
 data class CreateIdentitySessionRequest(
-    val refreshAddress: String,
     val requestAttributes: List<String>
```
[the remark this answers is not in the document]
pgurusinga: this is part of USEID-299

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
hannesschaletzky: +1

hannesschaletzky (approved):

mpanne (approved):

PR #53, opened 2022-08-12
Convert test properties to YAML

mpanne:
And add the needed test config and util functions for it.

### DIFF
--- a/src/test/kotlin/de/bund/digitalservice/useid/config/TestApplicationProperties.kt
+++ b/src/test/kotlin/de/bund/digitalservice/useid/config/TestApplicationProperties.kt
@@ -1,0 +1,17 @@
+package de.bund.digitalservice.useid.config
+
+import de.bund.digitalservice.useid.util.YamlPropertySourceFactory
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.annotation.PropertySource
+
+@EnableConfigurationProperties(TestApplicationProperties::class)
+@PropertySource("classpath:application.yaml", factory = YamlPropertySourceFactory::class)
+@ConfigurationProperties(prefix = "application")
+class TestApplicationProperties {
+    var staging: StagingProperties? = null
+
+    class StagingProperties {
+        lateinit var url: String
+    }
+}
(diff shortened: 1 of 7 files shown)


src/test/kotlin/de/bund/digitalservice/useid/config/TestApplicationProperties.kt
```diff
@@ -0,0 +1,17 @@
...
+import de.bund.digitalservice.useid.util.YamlPropertySourceFactory
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.annotation.PropertySource
+
+@EnableConfigurationProperties(TestApplicationProperties::class)
+@PropertySource("classpath:application.yaml", factory = YamlPropertySourceFactory::class)
+@ConfigurationProperties(prefix = "application")
+class TestApplicationProperties {
+    var staging: StagingProperties? = null
```
pgurusinga, line 12: why do we set the value as null? is it possible with `lateinit` as well?

pgurusinga (commented):
just a side note:
what do you think if we move all files to e.g `/YamlTestConfigSetup` package, since we won't touch all of these changes (A one-time code-config setup to enable Yaml).

src/test/kotlin/de/bund/digitalservice/useid/config/TestApplicationProperties.kt
mpanne: Somehow that leads to errors on initialisation. 
 
Screenshot 2022-08-15 at 10 13 41

mpanne: About your package idea: 
The YAML factory could be part of a "one-time" package which is not touched anymore. The TestApplicationProperties can be reused for any config that should be available in any tests so I would like to keep that in the config package.

pgurusinga: that would be great if we can separate that and also for the TestApplicationProperties, maybe we could add an information about the purpose of the file?

pgurusinga (approved):

src/test/kotlin/de/bund/digitalservice/useid/config/TestApplicationProperties.kt
mpanne: Like that?

hannesschaletzky (approved):

PR #66, opened 2022-08-31
UseID-231/get identity endpoint

hannesschaletzky:
In IdentificationSessionsController.kt, I have a reactive piece of code I could not solve in a good way... Please have a look. 

image

In doOnNext I need to call the service again to fetch the session. But it was fetched already in the first call ob the publisher list -> identificationSessionService.findByEIDSessionId(eIDSessionId). 
How can I keep a value and use it in a block later down the line? 
I've tried using [zipWith](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#zipWith-reactor.core.publisher.Mono-) and [Mono.zip](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#zip-java.util.function.Function-reactor.core.publisher.Mono...-) and [zipWhen](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#zipWhen-java.util.function.Function-), see tries of [commit](https://github.com/digitalservicebund/useid-backend-service/commit/71be398cffcbdc442e6757207c8f4e6c5f26948d) 'Refactor Suggestions', but that did not work. It returned an empty result. 

The logical chain should be as follows: 

1. check existence by fetching session
2. call eid server 
3. delete identificationSession if result from e id server is successful 
4. return response


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -11,15 +11,24 @@ class IdentificationSessionService(private val mockDatasource: MockDatasource) {
...
 
-    fun findById(useIDSessionId: UUID): Mono<IdentificationSession> {
-        return Mono.justOrEmpty(mockDatasource.findById(useIDSessionId))
+    fun findByUseIDSessionId(useIDSessionId: UUID): Mono<IdentificationSession> {
+        return Mono.justOrEmpty(mockDatasource.findByUseIDSessionId(useIDSessionId))
+    }
+
+    fun findByEIDSessionIdOrFail(eIDSessionId: UUID): IdentificationSession {
+        return mockDatasource.findByEIDSessionId(eIDSessionId)
+            ?: throw Error("no session found with eIDSessionId $eIDSessionId")
```
pgurusinga, line 24: shouldn't this throw `NoSuchElementException()` ?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -94,20 +97,40 @@ class IdentificationSessionsController(
...
-    fun getIdentity(@PathVariable eIDSessionId: UUID): Mono<IdentityAttributes> {
-        // Currently, mock identity
-        return Mono.just(IdentityAttributes("firstname", "lastname"))
-            .filter {
-                identificationSessionService.sessionExists(eIDSessionId)
+    fun getIdentity(@PathVariable eIDSessionId: UUID): Mono<ResponseEntity<GetResultResponseType>> {
+        return identificationSessionService.findByEIDSessionId(eIDSessionId)
+            .map {
+                eidService.getEidInformation(eIDSessionId.toString())
+            }
```
pgurusinga, line 111: since `getEidInformation` is (assuming) a blocking call, we could use `Mono.fromCallable` ?
https://betterprogramming.pub/how-to-avoid-blocking-in-reactive-java-757ec7024676

```kotlin
identificationSessionService.findByEIDSessionId(eIDSessionId)
.flatMap {
    Mono.fromCallable { eidService.getEidInformation(eIDSessionId.toString()) }
.subscribeOn(Schedulers.boundedElastic())
}
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -94,20 +97,40 @@ class IdentificationSessionsController(
...
+            .doOnNext {
+                // TODO: REPLACED CONTAINS CONDITION WITH FIXED SUCCESS CASE (e.g. http://www.bsi.bund.de/ecard/api/1.1/resultmajor#success) AS SOON AS WE HAVE E2E SUCCESS FLOW DONE
+                if (it.result.resultMajor.contains("api")) {
```
pgurusinga, line 114: why do we need to check the value of the result? did we decide to handle the error in our backend or let the eService do error handling?

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
```diff
@@ -43,4 +47,12 @@ class MockDatasource {
         session.eIDSessionId = eIDSessionId
         log.info { "set new eIDSessionId: ${session.eIDSessionId} for session with useIDSessionId: $useIDSessionId" }
     }
+
+    fun delete(session: IdentificationSession): Boolean {
```
pgurusinga, line 51: why do we return a `Boolean`? will the return value be consumed in the Controller? would it be ok to return `Unit` instead ?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -93,21 +96,40 @@ class IdentificationSessionsController(
...
             .doOnError { exception ->
-                log.error { "error occurred while getting the tc token for session with id $useIDSessionId; ${exception.message}" }
+                log.error { "error occurred while getting tc token for useIDSessionId $useIDSessionId;\n ${exception.message}" }
             }
             .onErrorReturn(
                 ResponseEntity.internalServerError().body(null)
             )
     }
 
     @GetMapping("/{eIDSessionId}")
```
pgurusinga, line 106: maybe we should extend the path?
`GET /api/v1/identification/sessions/{eIDSessionId}/identities`
`GET /api/v1/identification/sessions/{useIDSessionid}/identities/{eIDSessionId}`

Github has amazing public endpoints, maybe we could have it as a reference: https://docs.github.com/en/rest/overview/endpoints-available-for-github-apps

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
hannesschaletzky: We gotta decide on this one. 
Merlin and I said that it would be nice have different error cases to handle. 
Also, we only want to remove the session if the result was successfully given back; so we need at least a success case.

src/main/kotlin/de/bund/digitalservice/useid/identification/MockDatasource.kt
hannesschaletzky: What is unit about? I don't know that... 
I've also thought about throwing an exception when the element could not be removed. What you think? (So we would remove the return type entirely)

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
hannesschaletzky: Extracted this to new ticket: 
https://digitalservicebund.atlassian.net/browse/USEID-406

pgurusinga (approved):

PR #67, opened 2022-09-01
UseID-394 / Removed tcTokenService

hannesschaletzky:
[Ticket](https://digitalservicebund.atlassian.net/browse/USEID-394)

-> directly call SDK from controller. No need for a service


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -75,8 +76,8 @@ class IdentificationSessionsController(
     )
     fun getTCToken(@PathVariable useIDSessionId: UUID): Mono<ResponseEntity<TCTokenType>> {
         return identificationSessionService.findById(useIDSessionId)
-            .flatMap {
-                tcTokenService.getTcToken(it.refreshAddress)
+            .map {
+                eidService.getTcToken(it.refreshAddress)
             }
```
pgurusinga, line 81: I tried locally and still works fine...
```suggestion
            .flatMap {
                /*
                * We cannot just wrap a blocking call with a Mono or a Flux
                * https://betterprogramming.pub/how-to-avoid-blocking-in-reactive-java-757ec7024676
                */
                Mono.fromCallable {
                    eidService.getTcToken(it.refreshAddress) // a blocking operation
                }.subscribeOn(Schedulers.boundedElastic())
            }
```

pgurusinga (approved):
beside my comment

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
hannesschaletzky: flatmap extracts the value out of the mono. 
we can use map and remove the mono creation inside the block. 

will put the comment in, though.

PR #69, opened 2022-09-09
Use id 299/data groups

hannesschaletzky:
Data groups will be injected into the EidService through the constructor. No usage of WebFlux Context or SecurityContextHolder or RequestContextHolder... 
For this, the EidService is not a Bean anymore, but will be instantiated by the controller in the route. 
This has the very important advantage that each service is bound to a request, which in a a singleton class pattern could have leaded to potential problems when under high load the data groups will be written to a single class and not per request to an own instance. 

image

image

🚀🚀🚀🚀🚀


src/test/kotlin/de/bund/digitalservice/useid/eidservice/EidServiceTest.kt
```diff
@@ -13,13 +17,32 @@ import org.springframework.context.annotation.Import
...
+        assertEquals(operations.givenNames, required)
+        assertEquals(operations.nationality, required)
+    }
+
+    @Test
+    fun `getWebserviceRequest throws IllegalStateException when passed invalid list items as data groups`() {
+        val eidService = EidService(config, listOf("iAmInvalid", "meToo"))
+        Assertions.assertThrows(IllegalStateException::class.java) {
+            eidService.webserviceRequest.useOperations
+        }
```
pgurusinga, line 46: do we have a case where we use forbidden data groups? let's say GST uses DG1 and DG2 but actually they mistakenly use DG4 and DG5?

mpanne (approved):

src/test/kotlin/de/bund/digitalservice/useid/eidservice/EidServiceTest.kt
hannesschaletzky: This exception would only be thrown after the setup, since the data groups will not change per request. 
They will be statically fetched from the env vars.

pgurusinga (approved):

PR #73, opened 2022-09-14
USEID-434 Update design of the widget

pgurusinga:


src/main/kotlin/de/bund/digitalservice/useid/statics/WidgetController.kt
```diff
@@ -9,4 +9,7 @@ import reactor.core.publisher.Mono
 class WidgetController {
     @GetMapping("/widget")
     fun widget(model: Model): Mono<String> = Mono.just("widget")
+
+    @GetMapping("/incompatible")
```
pgurusinga, line 13: still unsure with the naming. `/not-supported` ?

hannesschaletzky: /not-supported sounds a bit smoother. But really unsure about this as well

src/main/resources/static/css/widget.css
```diff
@@ -119,9 +126,62 @@ h2 {
...
+    box-shadow:         0 0 0 4px #fff, 0 0 0 8px #004b76;
+}
 .identify h2 {
     margin: 0;
 }
 .widget {
     max-width: 400px;
 }
+
+/*INCOMPATIBLE PAGE*/
```
hannesschaletzky, line 141: Then you'd need to change it here and downwards as well.

hannesschaletzky (approved):
Looks good!

PR #76, opened 2022-09-16
Useid 311/apikey security layer get identity

hannesschaletzky:
`.onErrorReturn({ e -> e is SecurityException }, ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(null))`
In order for us to handle/catch different error types in the subscriber chain, we can add a predicate in `onErrorReturn` which lets us specify the return value according to the error thrown. 
Due to the fact that the route is secured with an APIKey, we can easily read the dataGroups through the `authentication.details as ApiKeyDetails` and have a smooth and sound solution without adding a new parameter into the `IdentificationSession`

Let me know what you guys think of the solution. Tried to test everything accordingly. 
Cheers & have a nice weekend 🚀

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
@@ -12,6 +12,7 @@ import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.http.server.reactive.ServerHttpRequest
 import org.springframework.security.core.Authentication
+import org.springframework.security.web.authentication.session.SessionAuthenticationException
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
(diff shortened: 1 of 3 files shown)


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
```diff
@@ -100,16 +100,22 @@ class IdentificationSessionsController(
...
         */
+        val apiKeyDetails = authentication.details as ApiKeyDetails
         val getIdentityResult = Mono.fromCallable {
             val eidService = EidService(config)
             eidService.getEidInformation(eIDSessionId.toString())
         }
         return identificationSessionService.findByEIDSessionId(eIDSessionId)
+            .doOnNext {
+                if (apiKeyDetails.refreshAddress != it.refreshAddress) {
+                    throw SecurityException("APIKey unauthorized")
```
mpanne, line 116: The `SecurityException` docs say "Thrown by the security manager to indicate a security violation." which is not really the case here. Hence, I would recommend to use Springs `AuthenticationException` which is "Abstract superclass for all exceptions related to an Authentication object being invalid for whatever reason." or use an appropriate subclass. Furthermore, I would suggest to add some more info to the exception message since the message will not be externalized but might help us when debugging or understanding the code.

```suggestion
                    throw AuthenticationException("API key differs from the API key used to start the identification session.")
```

mpanne (approved):
Nice and short solution. Like it!

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionsController.kt
hannesschaletzky: Yes. Been looking into the [AuthenticationException](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/core/AuthenticationException.html ) as well, but it is an abstract class and we cannot throw that with an individual message. 
So I figured that the [SessionAutheticationException](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/web/authentication/session/SessionAuthenticationException.html) can be instantiated and looks good for that use case.

PR #78, opened 2022-09-19
Add prometheus

phuesler:
In order to be able to track application metrics

```
curl localhost:8080/actuator/prometheus.
```

We will have to block access to the metrics on the ingress/api gateway or similar


pgurusinga (approved):

mpanne (approved):
Meanwhile two new controllers have been added where the `@Timed` annotation should be added. The `RefreshController` and the `WellKnownController`.

PR #79, opened 2022-09-20
USEID-392 Make refresh address passing same-origin-policy with subject URL

pgurusinga:
## Problem
Since the technical guideline describes, that refresh address should respect same origin policy, therefor we cannot use eService address directly.

## Proposed changes
1. Add `/refresh` endpoint as a proxy

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
@@ -19,10 +19,10 @@ class RefreshController(private val identificationSessionService: Identification
     private val log = KotlinLogging.logger {}
 
     @GetMapping
-    fun redirectService(@RequestParam sessionId: UUID): Mono<ResponseEntity<Unit>> {
-        return identificationSessionService.findByEIDSessionId(sessionId)
-            .doOnError { exception ->
-                log.error { "error occurred while checking $sessionId;\n ${exception.message}" }
+    fun redirectToEServiceRefreshAddress(@RequestParam("sessionId") eIDSessionId: UUID): Mono<ResponseEntity<Unit>> {
+        return identificationSessionService.findByEIDSessionId(eIDSessionId)
+            .doOnError {
+                log.error("Failed to load identification session with eIDSessionId: $eIDSessionId", it)
             }
             .map {
                 ResponseEntity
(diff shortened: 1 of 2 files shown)


src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
```diff
@@ -0,0 +1,38 @@
...
+class RefreshController(private val identificationSessionService: IdentificationSessionService) {
+
+    private val log = KotlinLogging.logger {}
+
+    @GetMapping
+    fun redirectService(@RequestParam sessionId: UUID): Mono<ResponseEntity<Unit>> {
+        return identificationSessionService.findByEIDSessionId(sessionId)
+            .doOnError { exception ->
+                log.error { "error occurred while checking $sessionId;\n ${exception.message}" }
+            }
```
mpanne, line 26: Even if it's a little longer that way, I would make the naming a little bit more explicit. And please log the whole exception, not only the exception message (by providing it to the log message as last parameter). Otherwise the stack trace will not be logged.  

```suggestion
    fun redirectToEServiceRefreshAddress(@RequestParam("sessionId") eIDSessionId: UUID): Mono<ResponseEntity<Unit>> {
        return identificationSessionService.findByEIDSessionId(eIDSessionId)
            .doOnError {
                log.error("Failed to load identification session with eIDSessionId: $eIDSessionId", it)
            }
```

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
```diff
@@ -0,0 +1,38 @@
...
+            .map {
+                ResponseEntity
+                    .status(HttpStatus.SEE_OTHER)
```
mpanne, line 29: 303 SEE OTHERS is "Used to redirect after a PUT or a POST, so that refreshing the result page doesn't re-trigger the operation." ([see here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections))
So I would suggest to use 301.

mpanne (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
pgurusinga: After discussion, we keep the 303 SEE OTHERS based on the behaviour of `redirect:` from Spring

mpanne (approved):

PR #80, opened 2022-09-22
USEID-479 Add endpoint for iOS universal link config file

pgurusinga:
## Problem
We want to enable Universal Link and we need to store it in our backend. The problem occurs (I don't know if this is an issue)  when the file is stored inside static resource, because the strict rule from Apple to not have the file extension, it causes the browser to download the file, because Spring assumes the file as non JSON and sets the file as `application/octet-stream`

## Proposed changes
Create a simple route for iOS

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
@@ -1,0 +1,38 @@
+package de.bund.digitalservice.useid.statics
+
+import mu.KotlinLogging
+import org.springframework.http.MediaType
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import reactor.core.publisher.Mono
+
+@Controller
+class DeeplinkController {
+
+    private val log = KotlinLogging.logger {}
+
+    @GetMapping(
+        path = [".well-known/apple-app-site-association"],
+        produces = [MediaType.APPLICATION_JSON_VALUE]
+    )
+    fun getIOSUniversalLink(): Mono<ResponseEntity<IOSUniversalLink>> {
+        val appIds = listOf(
+            "VDTVKQ35RL.de.bund.digitalservice.UseID",
+            "VDTVKQ35RL.de.bund.digitalservice.UseID-Preview"
+        )
+        val components = listOf(Component(pathUrlName = "/eID-Client"))
+        val details = listOf(Details(appIds, components))
+        val universalLink = UniversalLink(details)
+
+        return Mono.just(
+            ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(IOSUniversalLink(universalLink))
+        )
+            .doOnError {
+                log.error("Failed to return iOS Universal Link config", it)
+            }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        return Mono.just(
+            ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(IOSUniversalLink(universalLink))
+        )
+            .doOnError {
+                log.error("Failed to return iOS Universal Link config", it)
+            }
+            .onErrorReturn(
+                ResponseEntity.internalServerError().build()
+            )
```
pgurusinga, line 36: I don't see a way that would throw an error. 🤔

src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import mu.KotlinLogging
+import org.springframework.http.MediaType
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import reactor.core.publisher.Mono
+
+@Controller
+class DeeplinkController {
```
mpanne, line 11:
```suggestion
class WellKnownController {
```

src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    private val log = KotlinLogging.logger {}
+
+    @GetMapping(
+        path = [".well-known/apple-app-site-association"],
+        produces = [MediaType.APPLICATION_JSON_VALUE]
+    )
+    fun getIOSUniversalLink(): Mono<ResponseEntity<IOSUniversalLink>> {
```
mpanne, line 19:
```suggestion
    fun getAppleAppSiteAssociation(): Mono<ResponseEntity<IOSUniversalLink>> {
```

src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
```diff
@@ -0,0 +1,38 @@
...
+        val appIds = listOf(
+            "VDTVKQ35RL.de.bund.digitalservice.UseID",
+            "VDTVKQ35RL.de.bund.digitalservice.UseID-Preview"
+        )
+        val components = listOf(Component(pathUrlName = "/eID-Client"))
+        val details = listOf(Details(appIds, components))
+        val universalLink = UniversalLink(details)
```
mpanne, line 26: What if something changes here? That would mean we have to manually modify this code here?

mpanne (approved):

src/main/kotlin/de/bund/digitalservice/useid/statics/DeeplinkController.kt
pgurusinga: yes, it's the trade-off we have to face :( it's not a great developer experience but I see:
1. no read file system
2. we don't change this configuration that often

PR #83, opened 2022-09-27
Useid 401/configure postgres database access

mpanne:
We want to enhance the resilience and scalability of our backend by storing the identification sessions in a PostgreSQL database instead of in-memory. 

* Use Spring Data R2DBC to implement database connection
* Introduce Flyway as database migration tool


src/test/kotlin/de/bund/digitalservice/useid/refresh/RefreshControllerIntegrationTest.kt
```diff
@@ -20,8 +20,7 @@ import java.util.UUID
 private const val AUTHORIZATION_HEADER = "Bearer some-api-key"
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-@Tag("integration")
-class RefreshControllerIntegrationTest(@Autowired val webTestClient: WebTestClient) {
+class RefreshControllerIntegrationTest(@Autowired val webTestClient: WebTestClient) : PostgresTestcontainerIntegrationTest() {
```
pgurusinga, line 23: why does `RefreshControllerIntegrationTest` need to extend `PostgresTestcontainerIntegrationTest` ? is it because of the Controller uses repository?

pgurusinga (approved):

src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionRepositoryIntegrationTest.kt
```diff
@@ -0,0 +1,72 @@
+package de.bund.digitalservice.useid.identification
+
+import de.bund.digitalservice.useid.persistence.FlywayConfig
+import de.bund.digitalservice.useid.persistence.FlywayProperties
+import de.bund.digitalservice.useid.util.PostgresTestcontainerIntegrationTest
+import org.hamcrest.MatcherAssert.assertThat
+import org.hamcrest.Matchers.equalTo
+import org.hamcrest.Matchers.hasItems
+import org.hamcrest.Matchers.notNullValue
```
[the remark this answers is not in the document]
mpanne: Hamcrest is part of `spring-boot-starter-test`

src/test/kotlin/de/bund/digitalservice/useid/refresh/RefreshControllerIntegrationTest.kt
mpanne: The refresh endpoint uses the `IdentificationSessionService` which relies on the `IdentificationSessionRepository`. Furthermore, the test creates an identification session and fetches a tc token which are both operations that require the mentioned service and repository to work. Hence, either we need a test databse (e.g. by extending `PostgresTestcontainerIntegrationTest`) or we need to mock the service / repository.

src/main/resources/db/migration/V1__Create_Identification_Session_Table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE identification_session
+(
+    id SERIAL PRIMARY KEY,
+    useid_session_id UUID NOT NULL UNIQUE,
+    refresh_address VARCHAR NOT NULL,
+    request_data_groups VARCHAR(5)[] NOT NULL,
+    eid_session_id  UUID UNIQUE
+);
```
phuesler, line 8: I think we are missing the index to search for useid_session_id and eid_session_id

phuesler (changes requested):
We need to create an index in postgres for the two columns we use to find sessions by. Otherwise the lookup will result in a table scan. This won't matter much with only a few sessions, but can become a problem if the number of sessions increase

src/main/resources/db/migration/V1__Create_Identification_Session_Table.sql
mpanne: I thought about adding those indices but decided against it. The reason was that an index makes inserting and updating an entity slower whilst it speeds up searching for one. We do insert and update as often as we search for entities which is why I though we might not need indices. Furthermore, I read an article which said that adding an index on a column which contains a lot of null values (eidSessionId) is not recommended. Unfortunately it was not explained.

phuesler: Yes, updating of the values/columns that are in the index, which I don't think we do as much. Given this is a session storage and the amount of sessions is likely to be small < 1000 it probably doesn't matter too much at this point and we can always add indices later on. Can you please add a comment to the SQL file about the considerations, because I'm probably not the only one that will jump on this :-)

phuesler: Also, speed wise, everything will be in memory anyways, considering how small the data set is and how we sized the database for production. So I don't think there will be a noticeable performance difference in inserts and updates

mpanne: Ok but if it does not hurt, we could also just add indices instead of a comment.

mpanne: Seems like [PostreSQL automatically creates indices for columns with a unique constraints](https://www.postgresql.org/docs/8.1/sql-createtable.html#SQL-CREATETABLE-NOTES).
> PostgreSQL automatically creates an index for each unique constraint and primary key constraint to enforce uniqueness. Thus, it is not necessary to create an index explicitly for primary key columns. (See [CREATE INDEX](https://www.postgresql.org/docs/8.1/sql-createindex.html) for more information.)

phuesler (approved):

PR #84, opened 2022-09-28
USEID-473 Content Security Policy

pgurusinga:
# Problem 
X-Frame-Options currently only allows a website to be embeddable either SAME-ORIGIN or not embeddable at all. Content-Security-Policy `frame-ancestors` allows our page to be embedded on website that is listed on our end.

# Proposed Changes
- CSP is loaded in the Spring as application properties
- Spring Security sets CSP

### DIFF
--- a/src/test/kotlin/de/bund/digitalservice/useid/statics/WidgetControllerIntegrationTest.kt
+++ b/src/test/kotlin/de/bund/digitalservice/useid/statics/WidgetControllerIntegrationTest.kt
@@ -22,7 +22,7 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
     }
 
     @Test
-    fun `widget endpoint should deliver Content-Security-Policy with allowed host when the request URL is valid`() {
+    fun `widget endpoint returns Content-Security-Policy with allowed host when the request URL is valid`() {
         webTestClient
             .get()
             .uri("/widget?hostname=foo.bar")
@@ -37,10 +37,25 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
     }
 
     @Test
-    fun `widget endpoint should deliver default Content-Security-Policy when the request URL is invalid`() {
+    fun `widget endpoint returns default Content-Security-Policy when the request URL is invalid`() {
         webTestClient
             .get()
             .uri("/widget?hostname=not-allowed.com")
+            .exchange()
+            .expectStatus()
+            .isOk
+            .expectHeader()
+            .valueEquals(
+                "Content-Security-Policy",
+                "some default value;frame-ancestors 'self';"
+            )
+    }
+
+    @Test
+    fun `widget endpoint returns default Content-Security-Policy when query parameter is not set`() {
+        webTestClient
+            .get()
+            .uri("/widget")
             .exchange()
             .expectStatus()
             .isOk


pgurusinga:
@mpanne should we set the allowed website as environment variable or easiest, we hardcode in the `application.yaml` ?

src/main/resources/application-local.yaml
```diff
@@ -25,4 +25,8 @@ eidservice:
     keystore: classpath:testKeys/sign.localhost_2382.p12
     type: PKCS12
     alias: sign.localhost
-    password: 2382
\ No newline at end of file
+    password: 2382
+
+csp:
+  defaultConfig: "default-src 'self';script-src 'self';style-src 'self';font-src 'self';img-src 'self';connect-src 'self';"
+  frameAncestors: frame-ancestors 'self' localhost;
```
pgurusinga, line 32: The reason of the separation is to avoid mistake when editing a long string character

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -13,31 +13,32 @@ import org.springframework.security.web.server.authentication.ServerAuthenticati
...
-    ): SecurityWebFilterChain {
+    fun springSecurityFilterChain(http: ServerHttpSecurity): SecurityWebFilterChain {
         return http.authorizeExchange()
             .pathMatchers("$IDENTIFICATION_SESSIONS_BASE_PATH/*/tc-token").permitAll()
             .pathMatchers("$IDENTIFICATION_SESSIONS_BASE_PATH/**").authenticated()
             .anyExchange().permitAll()
             .and().csrf().disable()
-            .headers().frameOptions().disable()
+            .headers()
+            .contentSecurityPolicy(cspProperties.getCSPHeaderValue()).and()
```
mpanne, line 29: I assume this sets the CSP header for all endpoints. Don't we only need it for the `/widget` endpoint?

mpanne (changes requested):
I we add the list of registered domains in the properties file, it is public to the world and everyone knows, which customers we have and what domains are whitelisted. Therefore I would do it as env variable via the useid-infra repo.

Furthermore, we decided ([see adr](https://github.com/digitalservicebund/useid-backend-service/blob/9e7a090839452d21dd7451965b15339b6797ea12/doc/adr/0005-csp-for-embedding-widget.md)) that we only set one domain to the frame-ancestors directive dynamically depending on the calling domain to make sure the header is not getting too big and nobody can discover all whitelisted domains just by calling the endpoint. That is not covered in this PR yet, right?

mpanne (commented):
I think what we need could be something like the following. In the WidgetController adapt the `/widget` endpoint to add a response header depending on some domain parameter (pseudo code): 

```
@GetMapping("/widget")
fun widget(model: Model, response: ServerResponse, @RequestParam domain: String): Mono<String> {
    if (domainIsWhitelisted(domain)) {
        response.headers().set("Content-Security", "frame-ancestors $domain")
    }
    return Mono.just("widget")
}
```

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
pgurusinga: what I did earlier, I created 2 type security chains: `apiSpringSecurityFilterChain` and `webSpringSecurityFilterChain`, so that we can control from Spring Security rather Controller specific, in case we want to send frame-ancestors in controllers that have view e.g. `/incompatible`, but this is somehow not working as expected
Do you know where I made the mistake?

```kotlin
    fun apiSpringSecurityFilterChain(
        http: ServerHttpSecurity,
        authenticationManager: ReactiveAuthenticationManager,
        authenticationConverter: ServerAuthenticationConverter
    ): SecurityWebFilterChain {
        return http.authorizeExchange()
            .pathMatchers("$IDENTIFICATION_SESSIONS_BASE_PATH/*/tc-token").permitAll()
            .pathMatchers("$IDENTIFICATION_SESSIONS_BASE_PATH/**").authenticated()
            .anyExchange().permitAll()
            .and().addFilterAfter(
                apiAuthenticationFilter(authenticationManager, authenticationConverter),
                SecurityWebFiltersOrder.REACTOR_CONTEXT
            )
            .build()
    }
```

```kotlin
fun webSpringSecurityFilterChain(http: ServerHttpSecurity): SecurityWebFilterChain {

        return http.authorizeExchange()
            .pathMatchers("/widget").permitAll()
            .anyExchange().permitAll()
            .and().csrf().disable()
            .headers()
            .contentSecurityPolicy(contentSecurityPolicy.toString())
            .and()
            .frameOptions().disable()
            .and()
            .build()
```

mpanne: You inject the same `ServerHttpSecurity` singleton in both cases, so I guess one config overrides the other. And even if it just adds the config instead of overriding it, it would be applied to the same security chain so the headers would be applied to all endpoints. What you need is a config on controller or `pathMatcher` level.

pgurusinga: ok I guess for now let's do it more specific in the `/widget` controller

src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyProperties.kt
```diff
@@ -0,0 +1,28 @@
...
+    lateinit var frameAncestors: String
+
+    @NotEmpty
+    lateinit var allowedHosts: List<String>
+
+    fun domainIsAllowed(host: String): Boolean {
+        return allowedHosts.contains(host)
+    }
+    fun getCSPHeaderValue(host: String): String {
+        return "$defaultConfig$frameAncestors $host;"
```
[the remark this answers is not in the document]
pgurusinga: image

src/main/kotlin/de/bund/digitalservice/useid/statics/WidgetController.kt
```diff
@@ -1,15 +1,30 @@
...
 
 @Controller
-class WidgetController {
+class WidgetController(
+    private val contentSecurityPolicyProperties: ContentSecurityPolicyProperties
+) {
     @GetMapping("/widget")
-    fun widget(model: Model): Mono<String> = Mono.just("widget")
+    fun widget(serverHttpResponse: ServerHttpResponse, serverHttpRequest: ServerHttpRequest): Mono<String> {
+        val host = serverHttpRequest.uri.host
```
mpanne, line 16: This `host` is probably our domain, isn't it? Since it is the request URI. So this won't work. What we need is the domain of the origin of the request. So as described in [this ADR](https://github.com/digitalservicebund/useid-backend-service/blob/main/doc/adr/0005-csp-for-embedding-widget.md): "Therefore the widget.js file adds the domain of the parent page as query parameter to the iframe url."

src/main/kotlin/de/bund/digitalservice/useid/statics/WidgetController.kt
```diff
@@ -1,15 +1,30 @@
...
+        val allowedHost = contentSecurityPolicyProperties.domainIsAllowed(host)
+
+        if (allowedHost) {
+            serverHttpResponse.headers.set(
+                "Content-Security-Policy",
+                contentSecurityPolicyProperties.getCSPHeaderValue(host)
+            )
+        }
```
mpanne, line 24: Let's keep in mind to put all this to a custom filter in case it is needed for more than one endpoint at some point. For now it is ok to do it here though. 
Do the error pages of the widget need the same CSP header?

mpanne (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/statics/WidgetController.kt
pgurusinga: ah ok I missed that

src/main/kotlin/de/bund/digitalservice/useid/statics/WidgetController.kt
pgurusinga: hmm maybe with webfilter is much easier, but we cannot control per Controller, e.g. we need to define a path in the filter which can also lead to mistake... 
we could however make the path as constant...

mpanne: Yes, we could do that using constants. I think for now it's fine to have it in the Controller code. But let's keep in mind that we should move it to a Filter once we have to reuse it somewhere.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -1,15 +1,30 @@
...
+class WidgetController(
+    private val contentSecurityPolicyProperties: ContentSecurityPolicyProperties
+) {
     @GetMapping("/widget")
-    fun widget(model: Model): Mono<String> = Mono.just("widget")
+    fun widget(serverHttpResponse: ServerHttpResponse, serverHttpRequest: ServerHttpRequest): Mono<String> {
+        val host = serverHttpRequest.uri.host
+        val allowedHost = contentSecurityPolicyProperties.domainIsAllowed(host)
+
+        if (allowedHost) {
```
phuesler, line 19: What happens in the else case?

pgurusinga: I updated the logic, it nows filtered in `src/main/kotlin/de/bund/digitalservice/useid/widget/ContentSecurityPolicyFilter.kt` and when else, its response only returns `self` which only our domain can load the widget

src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyFilter.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override fun filter(
+        serverWebExchange: ServerWebExchange,
+        webFilterChain: WebFilterChain
+    ): Mono<Void> {
+        val request: ServerHttpRequest = serverWebExchange.request
+        val response: ServerHttpResponse = serverWebExchange.response
+        val pathIsValidWidgetPages = listOfPages.any { it.matches(request.path.pathWithinApplication()) }
+
+        if (pathIsValidWidgetPages) {
```
pgurusinga, line 30: @mpanne @hannesschaletzky I am not happy with the nested if statement, do you know how can we write this cleaner?

mpanne: You could just invert it: 

```
if (!pathIsValidWidgetPages) {
    return webFilterChain.filter(serverWebExchange)
}
```

src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyProperties.kt
```diff
@@ -0,0 +1,32 @@
...
```
mpanne, line 26: Even less code duplication ;) 

```suggestion
        return "${getDefaultCSPHeaderValue()} $host;"
```

src/test/kotlin/de/bund/digitalservice/useid/statics/WidgetControllerIntegrationTest.kt
```diff
@@ -10,14 +10,44 @@ import org.springframework.test.web.reactive.server.WebTestClient
...
+            .isOk
+            .expectHeader()
+            .valueEquals(
+                "Content-Security-Policy",
+                "some default value;frame-ancestors 'self' foo.bar;"
+            )
+    }
+
+    @Test
+    fun `widget endpoint should deliver default Content-Security-Policy when the request URL is invalid`() {
```
mpanne, line 40: Maybe we should add a similar test for the case that the parameter is not set at all?

src/test/resources/application.yaml
```diff
@@ -39,3 +39,7 @@ wellknown:
     package-preview:
       name: "de.digitalService.useID.Preview"
       fingerprint: "15:7E:42:A7:92:6B:A5:CA:9E:B0:29:8E:88:EE:81:0D:C3:13:E9:B5:84:41:50:28:8A:88:17:B4:14:40:FB:42"
+
+csp:
+  defaultConfig: "some default value;"
```
mpanne, line 44: Maybe time to add some default value ;)

mpanne (changes requested):
Does that work with the allowed hosts configured [here](https://github.com/digitalservicebund/useid-infra/blob/4b158740e69df71b1c01ed11cc3910c2d4f6bc0d/manifests/overlays/production/patch-deployment.yaml#L13-L20)? Is the parameter not only the host without protocol?

mpanne (approved):

src/test/resources/application.yaml
```diff
@@ -39,3 +39,10 @@ wellknown:
...
+  frameAncestors: "frame-ancestors 'self'"
+  allowedHosts:
+    - foo.bar
+    - bar.foo
```
mpanne, line 48: Those hosts should not be allowed on production or staging.

src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyProperties.kt
pgurusinga: we can add it because it will return `some default value;frame-ancestors 'self'; foo.bar;` the semicolon from default value is included, which should not happen.

PR #85, opened 2022-09-30
USEID-502 Add android app link association file

pgurusinga:
# Proposed changes
- make association strings as constants
- prettify the json output for better overview of what is being returned


src/main/kotlin/de/bund/digitalservice/useid/statics/WellKnownController.kt
```diff
@@ -7,6 +7,17 @@ import org.springframework.stereotype.Controller
...
+private const val IOS_APPID = "VDTVKQ35RL.de.bund.digitalservice.UseID"
+private const val IOS_APPID_PREVIEW = "VDTVKQ35RL.de.bund.digitalservice.UseID-Preview"
+private const val IOS_PATH_URL = "/eID-Client"
+
+private const val ANDROID_RELATION = "delegate_permission/common.handle_all_urls"
+private const val ANDROID_NAMESPACE = "android_app"
+private const val ANDROID_PACKAGE = "de.digitalService.useID"
+private const val ANDROID_PACKAGE_FINGERPRINT = "38:3A:63:60:60:BE:0B:E9:40:AB:F8:67:EA:BB:64:C8:91:99:0B:DC:01:D8:3D:34:89:A6:29:E4:1D:3B:85:9F"
+private const val ANDROID_PACKAGE_PREVIEW = "de.digitalService.useID.Preview"
+private const val ANDROID_PACKAGE_PREVIEW_FINGERPRINT = "15:7E:42:A7:92:6B:A5:CA:9E:B0:29:8E:88:EE:81:0D:C3:13:E9:B5:84:41:50:28:8A:88:17:B4:14:40:FB:42"
```
mpanne, line 19: I would suggest to move this to the `application.yaml` to separate the code from the actual properties.

src/main/kotlin/de/bund/digitalservice/useid/statics/WellKnownController.kt
```diff
@@ -7,6 +7,17 @@ import org.springframework.stereotype.Controller
...
+
 @Controller
 class WellKnownController {
```
mpanne, line 22: I don't really like that everything is in the `statics` package. What about having a `widget` package and a `well-known` package?

mpanne (approved):
Thanks for the fast work! 
Just some small suggestions.

src/main/kotlin/de/bund/digitalservice/useid/statics/WellKnownController.kt
pgurusinga: @mpanne I actually did that, because application properties is always loaded first and it's counted as reading a file, but I reverted it, since it's more complex for such a simple string definition, I had to create another class `WellKnownProperties`. I can also add that if that is more beneficial for us

src/main/kotlin/de/bund/digitalservice/useid/config/WellKnownProperties.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.stereotype.Component
+import org.springframework.validation.annotation.Validated
+import javax.validation.Valid
+import javax.validation.constraints.NotEmpty
+
+@Component
+@ConfigurationProperties(prefix = "wellknown")
+@Validated
+class WellKnownProperties {
```
mpanne, line 12: I like the new package structure. Why not move this to the `wellknown` package? :) It's not a global config but only needed in the package.

src/main/resources/application.yaml
```diff
@@ -32,4 +37,20 @@ eidservice:
...
+
+  android-config:
+    relation: "delegate_permission/common.handle_all_urls"
+    namespace: "android_app"
+    package-default:
+      name: "de.digitalService.useID"
+      fingerprint: "38:3A:63:60:60:BE:0B:E9:40:AB:F8:67:EA:BB:64:C8:91:99:0B:DC:01:D8:3D:34:89:A6:29:E4:1D:3B:85:9F"
+    package-preview:
+      name: "de.digitalService.useID.Preview"
+      fingerprint: "15:7E:42:A7:92:6B:A5:CA:9E:B0:29:8E:88:EE:81:0D:C3:13:E9:B5:84:41:50:28:8A:88:17:B4:14:40:FB:42"
```
mpanne, line 56: I like that.

mpanne (approved):

PR #90, opened 2022-10-06
USEID-UNPLANNED Add favicon

pgurusinga:
1. Safari throws error because missing favicon
image
2.  https://github.com/digitalservicebund/uptime-monitor/pull/70 for the icon


pgurusinga:
@phuesler thanks for the feedback, we will start with organizing the layout when we work on the fallback page.

PR #96, opened 2022-10-14
Useid 577/universal link fallback page

hannesschaletzky:
Added basic fallback page /eID-Client
Added Error Notification


src/main/resources/application-local.yaml
```diff
@@ -41,3 +41,4 @@ csp:
   allowed-hosts:
     - foo.bar
     - bar.foo
+    - localhost
```
hannesschaletzky, line 44: This could be removed.

src/main/resources/templates/components/errorNotification.mustache
```diff
@@ -0,0 +1,8 @@
+<div class="error_wrapper">
+    <img src="https://uploads-ssl.webflow.com/617fbee004d9ae0e43297700/6347f081beb57e1344756db2_img_icon_error_filler_red.svg" loading="lazy" alt="" class="error_icon">
```
mpanne, line 2: Please provide the image via our static directory instead of loading it from the internet. We do not control this link, so it can change which would break our view. Furthermore, loading the image via link directly to the users device makes it exploitable for delivery chain attacks.

src/main/resources/templates/components/errorNotification.mustache
```diff
@@ -0,0 +1,8 @@
...
+    <div>
+        <div class="error_title">Es ist ein Fehler aufgetreten.</div>
+        <div>Stellen Sie sicher, dass die BundesIdent App installiert ist und versuchen Sie es erneut.</div>
```
mpanne, line 5: Since we now store all the plain text of the widget in our `application.yaml` we should be consistent and also move this text there.

mpanne (changes requested):

mpanne (approved):

PR #101, opened 2022-10-20
Add link to data privacy statement to widget

mpanne:


hannesschaletzky (changes requested):
Build failed. Data Privacy Button default message seems to be empty... 

image

PR #104, opened 2022-10-24
Useid 578/widget tracking

hannesschaletzky:
Calls are fired and you can check the events on [Matomo Live](https://bund.matomo.cloud/index.php?module=CoreHome&action=index&idSite=2&period=range&date=last30#?idSite=2&period=range&date=last30&category=General_Visitors&subcategory=General_RealTime) and [Matomo Events](https://bund.matomo.cloud/index.php?module=CoreHome&action=index&idSite=2&period=range&date=last30#?idSite=2&period=range&date=last30&category=General_Actions&subcategory=Events_Events). Widget also tracks Button Pressed event and calls backend route /app-openend. 

image

Implementing the service and the matomo events, I'v realized that we are not having the WidgetController Routes in Reactive Style. Should I rework these routes to be reactive? 
I would vote for yes. (but I'd do this directly on main. PRs suck)
Please let me know @mpanne @code28 

Also I am unsure about how much testing the tracking needs... Please share your thoughts about it!

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingService.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingService.kt
@@ -1,0 +1,35 @@
+package de.bund.digitalservice.useid.tracking
+
+import mu.KotlinLogging
+import org.springframework.stereotype.Service
+import java.net.URI
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
+
+@Service
+class TrackingService(private val trackingProperties: TrackingProperties) {
+
+    private val log = KotlinLogging.logger {}
+    private val httpClient: HttpClient = HttpClient.newBuilder().build()
+    private val responseHandler: HttpResponse.BodyHandler<String> = HttpResponse.BodyHandlers.ofString()
+
+    private fun sendTrackingRequest(url: String) {
+        val request = HttpRequest.newBuilder()
+            .uri(URI.create(url))
+            .build()
+
+        val response = httpClient.send(request, responseHandler)
+        val status = response.statusCode()
+        if (status != 200) {
+            log.error("$status, tracking failed for: $url")
+        }
+        log.info("$status, successfully tracked: $url")
+    }
+
+    fun sendMatomoEvent(category: String, action: String, name: String) {
+        val matomoSiteId = trackingProperties.matomoSiteId
+        val domain = trackingProperties.matomoDomain
+        sendTrackingRequest("$domain?idsite=$matomoSiteId&rec=1&ca=1&e_c=$category&e_a=$action&e_n=$name")
+    }
+}
--- a/src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
@@ -1,0 +1,29 @@
+package de.bund.digitalservice.useid.tracking
+
+import org.springframework.stereotype.Component
+
+@Component
+class WidgetTracking {
+
+    class Categories {
+        val widget = "Widget"
+    }
+
+    class Actions {
+        val loaded = "loaded"
+        val buttonPressed = "buttonPressed"
+        val error = "error"
+    }
+
+    class Names {
+        val widget = "widget"
+        val startIdent = "startIdent"
+        val incompatible = "incompatible"
+        val unsupported_os = "unsupported_os"
+        val fallback = "fallback"
+    }
+
+    val categories = Categories()
+    val actions = Actions()
+    val names = Names()
+}
--- a/src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
@@ -19,7 +19,7 @@ class WidgetTracking {
         val widget = "widget"
         val startIdent = "startIdent"
         val incompatible = "incompatible"
-        val unsupported_os = "unsupported_os"
+        val unsupportedOS = "unsupportedOS"
         val fallback = "fallback"
     }
 
--- a/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
@@ -1,6 +1,8 @@
 package de.bund.digitalservice.useid.widget
 
 import de.bund.digitalservice.useid.config.ApplicationProperties
+import de.bund.digitalservice.useid.tracking.TrackingService
+import de.bund.digitalservice.useid.tracking.WidgetTracking
 import io.micrometer.core.annotation.Timed
 import org.springframework.http.HttpStatus
 import org.springframework.http.server.reactive.ServerHttpRequest
@@ -17,7 +19,9 @@ internal const val FALLBACK_PAGE = "eID-Client"
 @Timed
 class WidgetController(
     private val applicationProperties: ApplicationProperties,
-    private val widgetProperties: WidgetProperties
+    private val widgetProperties: WidgetProperties,
+    private val trackingService: TrackingService,
+    private val widgetTracking: WidgetTracking
 ) {
     private val defaultViewHeaderConfig = mapOf(
         "baseUrl" to applicationProperties.baseUrl,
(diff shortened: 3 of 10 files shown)


src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingService.kt
```diff
@@ -0,0 +1,35 @@
...
+        if (status != 200) {
+            log.error("$status, tracking failed for: $url")
+        }
+        log.info("$status, successfully tracked: $url")
+    }
+
+    fun sendMatomoEvent(category: String, action: String, name: String) {
+        val matomoSiteId = trackingProperties.matomoSiteId
+        val domain = trackingProperties.matomoDomain
+        sendTrackingRequest("$domain?idsite=$matomoSiteId&rec=1&ca=1&e_c=$category&e_a=$action&e_n=$name")
```
code28, line 33: Are we only tracking events? Weren't three of them page views?
And maybe we also want to track user agent or language?
Maybe Max can clarify, what is best, and answer those questions.

src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
```diff
@@ -0,0 +1,29 @@
...
+        val loaded = "loaded"
+        val buttonPressed = "buttonPressed"
+        val error = "error"
+    }
+
+    class Names {
+        val widget = "widget"
+        val startIdent = "startIdent"
+        val incompatible = "incompatible"
+        val unsupported_os = "unsupported_os"
```
code28, line 22: camelCase and snake_case mixed?
If wished like that in Matomo, I'd say we should at least only use one in our code.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -1,31 +1,54 @@
...
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.reactive.result.view.Rendering
 
 internal const val WIDGET_PAGE = "widget"
 internal const val INCOMPATIBLE_PAGE = "incompatible"
 internal const val FALLBACK_PAGE = "eID-Client"
+internal const val APP_OPENED = "app-opened"
```
code28, line 19: I know I suggested that name, but having a second thought, maybe "button clicked" is better?

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -1,31 +1,54 @@
...
+    private val widgetTracking: WidgetTracking
 ) {
     private val defaultViewHeaderConfig = mapOf(
         "baseUrl" to applicationProperties.baseUrl,
         "metaTag" to widgetProperties.metaTag
     )
 
+    @PostMapping("/$APP_OPENED")
+    fun handleAppOpened(): ResponseEntity<String> {
+        trackingService.sendMatomoEvent(
```
code28, line 36: I'm not that much into the spring/kotlin details, but is this always working? Don't we have to add this into a background task? Or is that done automatically?

src/main/resources/static/js/attachButtonTracking.js
```diff
@@ -0,0 +1,9 @@
+document
```
code28, line 1: Maybe rename this file to `attachAppButtonOnclick`?

code28 (changes requested):
Nice!
For reactive rework: I'd also vote for yes, but PR instead of directly on main 🤓
For testing: The integration test should at least check, that the `TrackingService` is called with the correct parameters.

src/main/kotlin/de/bund/digitalservice/useid/tracking/WidgetTracking.kt
chrkaatz: I agree

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
mpanne: I agree. `app-opened` is not very self explanatory without knowing / understanding the context of the widget.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
mpanne: What do you mean with "background task"?

src/main/resources/static/js/attachButtonTracking.js
```diff
@@ -0,0 +1,9 @@
...
+  .getElementById("eid-client-button")
+  .addEventListener("click", async function () {
+    await fetch("/app-opened", {
+      cache: "no-store",
+      method: "POST",
+      keepalive: true,
+    });
```
mpanne, line 8: What about csrf protection?

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
code28: I have to admit, I might be mixing sth up, but the call to matomo should be asynchronous. I don't remember at the moment how this is handled in Spring.

mpanne (changes requested):
Nicely done! 
I would also vote for reactive refactoring in a **new** PR.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -1,31 +1,54 @@
...
+        return ResponseEntity.status(HttpStatus.OK).body("")
+    }
+
     @GetMapping("/$WIDGET_PAGE")
     fun getWidgetPage(model: Model): Rendering {
+        trackingService.sendMatomoEvent(
+            widgetTracking.categories.widget,
+            widgetTracking.actions.loaded,
+            widgetTracking.names.widget
+        )
```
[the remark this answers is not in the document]
mpanne: Though about it again. If this controller is refactored to reactive way, this is almost implicitly solved. So ignore this comment for now but keep it in mind for the refactoring.

src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingService.kt
hannesschaletzky: Sent this to Max and he will get back to us. 
However, we are not having any tickets for tracking more things than events. 
I would vote for us to track events first and increase tracking activity with a new ticket.

maximilianwoltersdorf: We would for now only track events, not pageview. We can convert events into goals in matomo. that would allow us to to have events, as often as they occur, but goals only once per user. Which is a great advantage because this allows to answer the question "how many users struggled" or "how many users succeeded" way more accurately. I hope that makes sense?

maximilianwoltersdorf: user agent and language are for now not of interest since we only identified the few main KPIs. We try to work according to the principle to only track the must haves for now. anything else will be subject for improvements since we do not want to get any negative feedback from BSI, BMI or the community. Any dimension we identify along the road must be explicitly requested with a clear motivation for tracking

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
code28: > We should definitely execute the sending event asynchronously and immediately respond with OK. But that might be solved implicitly if the controller is refactored to reactive approach? Is that what you talking about @code28 ?

Yes :)

src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingService.kt
code28: For me it doesn't matter. And if you say, we track events, then events it is! :) 

> user agent and language are for now not of interest since we only identified the few main KPIs. We try to work according to the principle to only track the must haves for now. anything else will be subject for improvements since we do not want to get any negative feedback from BSI, BMI or the community. Any dimension we identify along the road must be explicitly requested with a clear motivation for tracking

Really nice, fully agree!

hannesschaletzky:
Thanks for your feedback, peeps!

PR #116, opened 2022-11-01
Add padding to widget fallback page

mpanne:
Add some padding to the fallback widget page since it is always opened in a new tab without a surrounding container.

### DIFF
--- a/src/main/resources/static/css/widget.css
+++ b/src/main/resources/static/css/widget.css
@@ -163,7 +163,7 @@ h2 {
     max-width: 400px;
 }
 .widget.fallback {
-    padding: 1rem;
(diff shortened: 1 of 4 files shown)


src/main/resources/static/css/widget.css
```diff
@@ -162,6 +162,9 @@ h2 {
 .widget {
     max-width: 400px;
 }
+.widget.fallback {
+    padding: 1rem;
```
jerdesign, line 166: Instead of REM let's use 4VW for the padding as REM is a font size based unit.

mpanne: Done

PR #118, opened 2022-11-02
USEID-617_refactor_tracking_service

hannesschaletzky:
Tracking Events are disabled for local profile. The class MatomoTrackingService will only listen on all profiles that are not local, e.g. Test, Staging, Production. To enable, remove the @Profile('!local') in the trackingService and the trackingProperties.

To create a new tracking service:

create new tracking event(s)
create a new tracking service that listens for the event(s)
shoot events from anywhere in the app
If we want to track more than just the event in Matomo:
a. create new event
b. add listener in MatomoTrackingService for that event
c execute api call with newly constructed URL

Please share your thoughts!


src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    private fun sendEvent(e: MatomoEvent) {
+        Mono.fromCallable {
+            constructEventURL(e)
+        }.zipWhen {
+            webRequests.POST(it)
+        }.map {
+            val status = it.t2.statusCode
+            if (status == HttpStatus.OK) {
+                log.info("$status, successfully tracked: ${it.t1}")
```
code28, line 44: As already mentioned, I find `info` a bit too high. But maybe that's for another ticket throughout the whole project.

code28 (approved):
Nice, LGTM!

src/main/kotlin/de/bund/digitalservice/useid/tracking/WebRequests.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun POST(url: String): Mono<ResponseEntity<Void>> {
+        return client
+            .post()
+            .uri(URI(url))
+            .retrieve()
+            .toBodilessEntity()
+            .onErrorReturn(
+                // everything but 200 will be caught and responded with 500
+                ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(null)
+            )
```
mpanne, line 22: Why do we need this? The response which is returned to the caller is only checked for success (200 OK) anyways. Can be removed imho.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -106,11 +107,16 @@ class WidgetController(
 
         return Rendering
             .view(WIDGET_PAGE)
-            .model(defaultViewHeaderConfig + widgetViewConfig)
+            .model(defaultViewHeaderConfig + widgetViewFallbackConfig)
             .status(HttpStatus.OK)
             .build()
     }
 
+    private fun publishMatomoEvent(category: String, action: String, name: String) {
```
mpanne, line 115: Category is always "Widget" since it's the widget controller. So this could be inlined in the method and the method could have a name like `publishWidgetMatomoEvent`.

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -0,0 +1,52 @@
...
+            } else {
+                log.error("$status, tracking failed for: ${it.t1}")
+            }
+        }.subscribe()
```
mpanne, line 48: Add error handling for errors in event stream (`onError`). Currently only API responses with status code not 200 are handled. And is a log message enough? Do we have an alarming for too many error logs (general question, not only for tracking)?

src/test/kotlin/de/bund/digitalservice/useid/tracking/WebRequestsTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.jupiter.api.Tag
+import org.junit.jupiter.api.Test
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.reactive.function.client.WebClient
+import reactor.core.publisher.Mono
+import java.net.URI
+
+@Tag("test")
+class WebRequestsTest {
```
mpanne, line 17: I like how you cleaned up this test! Nice!

src/test/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingServiceTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.system.CapturedOutput
+import org.springframework.boot.test.system.OutputCaptureExtension
+import org.springframework.context.ApplicationEventPublisher
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import reactor.core.publisher.Mono
+
+@ExtendWith(value = [OutputCaptureExtension::class, SpringExtension::class])
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class MatomoTrackingServiceTest : PostgresTestcontainerIntegrationTest() {
```
mpanne, line 22:
```suggestion
class MatomoTrackingServiceIntegrationTest : PostgresTestcontainerIntegrationTest() {
```

src/test/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingServiceTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        assertEquals(expectedURL, url)
+    }
+
+    @Test
+    fun `matomo tracking service should trigger web request and log event category, action and name and code 200`(output: CapturedOutput) {
+        val matomoEvent = MatomoEvent(this, "log1", "log2", "log3")
+        applicationEventPublisher.publishEvent(matomoEvent)
+        every { webRequests.POST(any()) } returns Mono.empty()
+        verify { webRequests.POST(any()) }
+    }
```
mpanne, line 53: I would love if we could stick to the GIVEN, WHEN, THEN order 😇

mpanne (approved):
Love the refactoring and like the event based approach! 

If you would ask me, I would even further decouple the `widget` package from the `tracking` package bei not sending `MatomoEvents` but publishing e.g. `WidgetLoadedEvent` and listen to those in the `MatomoTrackingService`. But that's for another time and to be discussed. :)

src/main/kotlin/de/bund/digitalservice/useid/tracking/WebRequests.kt
phuesler: Regardless of what the caller does, imo the function should let the caller know if something went wrong

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
phuesler: I think the log level on production is error anyway, so we should be ok.

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
phuesler: :+1 We have have prometheus metrics now. I think this services should be instrumented in general. Especially since it talks to an external endpoint. I think this is an extra ticket.

src/main/kotlin/de/bund/digitalservice/useid/tracking/WebRequests.kt
mpanne: If we just return the response which we received from the API, the caller can handle it in the desired way. Why mapping it here (and losing information)?

PR #121, opened 2022-11-03
USEID-377 monitor eid service

phuesler:
Add counters and timers to the eid service calls. I don't know if there is a better way to annotate inherited methods other than overriding and calling super.

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/eidservice/EidService.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/eidservice/EidService.kt
@@ -1,10 +1,13 @@
 package de.bund.digitalservice.useid.eidservice
 
 import de.bund.bsi.eid230.AttributeRequestType
+import de.bund.bsi.eid230.GetResultResponseType
 import de.bund.bsi.eid230.OperationsRequestorType
 import de.bund.bsi.eid230.UseIDRequestType
 import de.governikus.autent.sdk.eidservice.config.EidServiceConfiguration
 import de.governikus.autent.sdk.eidservice.eidservices.EidService230
+import de.governikus.autent.sdk.eidservice.tctoken.TCTokenType
+import io.micrometer.core.annotation.Timed
 
 /*
     ("Why do we need to override getWebserviceRequest")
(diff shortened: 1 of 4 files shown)


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -36,6 +38,11 @@ class IdentificationSessionsController(
     private val eidServiceConfig: EidServiceConfiguration
 ) {
     private val log = KotlinLogging.logger {}
+    private val tcTokenCallsSuccessfulCounter: Counter =
+        Metrics.counter("eid_service.requests", "method", "get_tc_token", "status", "200")
+    private val tcTokenCallsWithErrorsCounter: Counter = Metrics.counter("eid_service.requests", "method", "get_tc_token", "status", "500")
+    private val getEidInformationCallsSuccessfulCounter: Counter = Metrics.counter("eid_service.requests", "method", "get_eid_information", "status", "200")
+    private val getEidInformationCallsWithErrorsCounter: Counter = Metrics.counter("eid_service.requests", "method", "get_eid_information", "status", "200")
```
code28, line 45:
```suggestion
    private val getEidInformationCallsWithErrorsCounter: Counter = Metrics.counter("eid_service.requests", "method", "get_eid_information", "status", "500")
```

src/main/kotlin/de/bund/digitalservice/useid/eidservice/EidService.kt
```diff
@@ -60,4 +63,14 @@ class EidService constructor(config: EidServiceConfiguration, private val dataGr
         request.useOperations = selector
         return request
     }
+
+    @Timed(value = "eid_service.get_tc_token")
+    public override fun getTcToken(refreshUrl: String?): TCTokenType {
+        return super.getTcToken(refreshUrl)
+    }
+
+    @Timed(value = "eid_service.get_eid_infirmation")
```
code28, line 72:
```suggestion
    @Timed(value = "eid_service.get_eid_information")
```

mpanne (approved):

hannesschaletzky (approved):
Does it make sense to extract the "Metrics" into an own service like we did with the tracking? IdentificationSessionController.kt feels a little packed.

phuesler:
I agree that the controller contains a lot of code and is packed. However, I think the service communication part should be extracted into its own thing to slim down the controller. The success and response counting should happen where we handle the responses, the call counting is done by the timed annotations.

PR #125, opened 2022-11-14
USEID-625: Add load tests for 6 sessions per sec

code28:
Our goal was to check, if at least 5 requests per second work, so this load test checks for 6 sessions per second to be safe.

### DIFF
--- a/tests/load/README.md
+++ b/tests/load/README.md
@@ -8,6 +8,16 @@ brew install k6
 
 ## Running tests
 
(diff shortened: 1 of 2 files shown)


mpanne (approved):
Looks great! Thanks! 
As discussed it would be awesome if the URL and API key could somehow be configured without adapting the script. 
Furthermore, if running against the application running on localhost the VU limit is exceeded and the session duration threshold too. Maybe the limits should somehow be adapted?

PR #129, opened 2022-11-17
USEID-572: incompatible OS page info

pgurusinga:
# Problem
User cannot see directly which OS version is supported

# Proposed changes
1. Add user agent library `com.github.ua-parser:uap-java:1.5.3` . it is lightweight and the API for OS version is simpler than other candidate. This library breaks down the versioning into 3 properties (major, minor and patch) where other libraries do not.
2. Safely convert the version from the parser, since the parser will return a String, we need to transform the String into an Integer so that we can perform logical operation (>= or <=), so we need to safely cast it, in case of missing version number or unknown OS version, we show the widget instead (still need a product decision) 
3. Add 4 test cases where the string of
- User agent is correct
- User agent has no version
- User agent is incorrect
- User agent with unsupported version
4. Fix the incompatible page. The top area is out of view. 
image

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
@@ -146,12 +146,6 @@ class WidgetController(
         return "eidClientURL" to url
     }
 
-    /*
-       Since user agent will return version in String, we need to properly convert the string to integer,
-       otherwise we cannot perform logical operations such as <= or >=
-       If major version returns empty or somehow the user agent parser cannot parse the version properly,
-       we will show to users the widget
-    */
 
     private fun isIncompatibleOSVersion(userAgent: String): Boolean {
         val parsedUserAgent = Parser().parse(userAgent)
--- a/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
@@ -13,6 +13,7 @@ import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.reactive.result.view.Rendering
+import ua_parser.Client
 import ua_parser.Parser
 
 internal const val WIDGET_PAGE = "widget"
@@ -63,8 +64,8 @@ class WidgetController(
         )
 
         val parsedUserAgent = Parser().parse(userAgent)
-        val incompatibleIOSVersion = parsedUserAgent.os.family == "iOS" && safelyGetMajorOSVersion(parsedUserAgent.os.major) <= 14
-        val incompatibleAndroidVersion = parsedUserAgent.os.family == "Android" && safelyGetMajorOSVersion(parsedUserAgent.os.major) <= 8
+        val incompatibleIOSVersion = hasSupportedMajorOSVersion(parsedUserAgent, "iOS", 14)
+        val incompatibleAndroidVersion = hasSupportedMajorOSVersion(parsedUserAgent, "Android", 8)
 
         if (incompatibleIOSVersion || incompatibleAndroidVersion) {
             return Rendering
@@ -149,13 +150,17 @@ class WidgetController(
        If major version returns empty or somehow the user agent parser cannot parse the version properly,
        we will show to users the widget
     */
-    private fun safelyGetMajorOSVersion(majorVersion: String?): Int {
-        val fallbackOSVersion = 20
-
+    private fun safelyGetMajorOSVersion(majorVersion: String?): Int? {
         return try {
             Integer.parseInt(majorVersion)
         } catch (exception: NumberFormatException) {
-            fallbackOSVersion
+            null
         }
+    }
+
+    private fun hasSupportedMajorOSVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
+        return parsedUserAgent.os.family == osFamily &&
+            parsedUserAgent.os.major != null &&
+            safelyGetMajorOSVersion(parsedUserAgent.os.major)!! <= supportedMajorVersion
     }
 }
--- a/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
@@ -155,8 +155,8 @@ class WidgetController(
 
     private fun isIncompatibleOSVersion(userAgent: String): Boolean {
         val parsedUserAgent = Parser().parse(userAgent)
-        val incompatibleIOSVersion = hasIncompatibleMajorVersion(parsedUserAgent, "iOS", 14)
-        val incompatibleAndroidVersion = hasIncompatibleMajorVersion(parsedUserAgent, "Android", 8)
+        val incompatibleIOSVersion = hasIncompatibleMajorVersion(parsedUserAgent, "iOS", 15)
+        val incompatibleAndroidVersion = hasIncompatibleMajorVersion(parsedUserAgent, "Android", 9)
 
         return incompatibleIOSVersion || incompatibleAndroidVersion
     }
@@ -164,6 +164,6 @@ class WidgetController(
     private fun hasIncompatibleMajorVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
         return parsedUserAgent.os.family == osFamily &&
             !parsedUserAgent.os.major.isNullOrEmpty() &&
-            Integer.parseInt(parsedUserAgent.os.major) <= supportedMajorVersion
+            Integer.parseInt(parsedUserAgent.os.major) < supportedMajorVersion
     }
 }
--- a/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
@@ -66,8 +66,9 @@ class WidgetController(
         val parsedUserAgent = Parser().parse(userAgent)
         val incompatibleIOSVersion = hasIncompatibleMajorVersion(parsedUserAgent, "iOS", 14)
         val incompatibleAndroidVersion = hasIncompatibleMajorVersion(parsedUserAgent, "Android", 8)
+        val isIncompatibleOSVersion = incompatibleIOSVersion || incompatibleAndroidVersion
 
-        if (incompatibleIOSVersion || incompatibleAndroidVersion) {
+        if (isIncompatibleOSVersion) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -127,4 +142,20 @@ class WidgetController(
...
     }
+
+    /*
+       Since user agent will return version in String, we need to properly convert the string to integer,
+       otherwise we cannot perform logical operations such as <= or >=
+       If major version returns empty or somehow the user agent parser cannot parse the version properly,
+       we will show to users the widget
+    */
+    private fun safelyGetMajorOSVersion(majorVersion: String?): Int {
+        val fallbackOSVersion = 20
```
code28, line 153: Magic numbers aren't a good pattern, if they don't stand for what they mean (i.e. 20 would be only okay here, if `majorVersion` is always empty exactly on version 20).
Therefore, rather return `null` here (and `Int?` as return type) and e.g. add another check `!= null` in rows 66 and 67.

src/main/resources/static/css/widget.css
```diff
@@ -167,6 +167,9 @@ h2 {
 }
 
 /*INCOMPATIBLE PAGE*/
+.incompatible {
+    margin-top: 65px;
```
code28, line 171: Good catch! Actually, 65 is 1 px too much, but I think that's fine. 😉

code28 (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
pgurusinga: yes, I agree! I will try the return type! thanks!

src/main/resources/static/css/widget.css
pgurusinga: yes I did initially 64px, but somehow in my 1080p external screen, it looks like the top edge is cut off, so that's why I added 65px 😅

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
pgurusinga: done! please refer to this commit: https://github.com/digitalservicebund/useid-backend-service/pull/129/commits/0b5da7389bfd1560a4cf878edf554300b8b92a68

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -149,13 +150,17 @@ class WidgetController(
...
         } catch (exception: NumberFormatException) {
-            fallbackOSVersion
+            null
         }
     }
+
+    private fun hasSupportedMajorOSVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
+        return parsedUserAgent.os.family == osFamily &&
+            parsedUserAgent.os.major != null &&
+            safelyGetMajorOSVersion(parsedUserAgent.os.major)!! <= supportedMajorVersion
```
code28, line 164: This needs to be the other way round, right? We support versions "from 14" / "from 8" upwards and not downwards.

code28 (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -127,4 +143,24 @@ class WidgetController(
...
+            Integer.parseInt(majorVersion)
+        } catch (exception: NumberFormatException) {
+            null
+        }
+    }
+
+    private fun hasSupportedMajorOSVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
+        return parsedUserAgent.os.family == osFamily &&
+            parsedUserAgent.os.major != null &&
+            safelyGetMajorOSVersion(parsedUserAgent.os.major)!! <= supportedMajorVersion
```
mpanne, line 164: If `safelyGetMajorOSVersion` returns `null` this will throw an exception which will lead to an 500 error response to the user. This must be handled gracefully.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
mpanne: The name of the method is misleading. The method is used to set the boolean values `incompatibleIOSVersion` and  `incompatibleAndroidVersion`. So the method  returning `true` means that the version is incompatible. Please rename the method or reverse the logic.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -57,6 +63,16 @@ class WidgetController(
             "additionalClass" to ""
         )
 
+        val parsedUserAgent = Parser().parse(userAgent)
+        val incompatibleIOSVersion = hasSupportedMajorOSVersion(parsedUserAgent, "iOS", 14)
+        val incompatibleAndroidVersion = hasSupportedMajorOSVersion(parsedUserAgent, "Android", 8)
+
+        if (incompatibleIOSVersion || incompatibleAndroidVersion) {
```
mpanne, line 70: Just a matter of taste but this could be extracted to a method like `isOSSupported()` which would keep the handler method cleaner.

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -71,6 +71,98 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
...
+        val compatibleAndroidUserAgent = "Mozilla/5.0 (Linux; Android 10) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.5304.105 Mobile Safari/537.36"
+        val compatibleIOSUserAgent = "Mozilla/5.0 (iPhone; CPU iPhone OS 16_1_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.1 Mobile/15E148 Safari/604.1"
+
+        webTestClient
+            .get()
+            .uri("/widget")
+            .header("User-Agent", compatibleAndroidUserAgent)
+            .exchange()
+            .expectStatus()
+            .isOk
```
mpanne, line 85: Several times the same request. Could be extracted to a method.

mpanne (changes requested):
Nice! Some cleanup needed though and the widget shouldn't throw an exception in case the version cannot be extracted from the user agent.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
code28: 🙃

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
pgurusinga: so, `parsedUserAgent.os.major` will always return value of type `String` in case the the user agent is correct or wrong (empty string for invalid or missing).  
Therefor I simplify the logic for checking the major version, also I will remove `safelyGetMajorOSVersion` which I will explain at the bottom.

```kotlin
private fun hasIncompatibleMajorVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
        return parsedUserAgent.os.family == osFamily &&
            !parsedUserAgent.os.major.isNullOrEmpty() &&
            Integer.parseInt(parsedUserAgent.os.major) <= supportedMajorVersion
    }
```

since we are using double ampersand `&&` operator, if the first left statement false, it won't evaluate the right side (short-circuit evaluation).

```kotlin
!parsedUserAgent.os.major.isNullOrEmpty() &&
            Integer.parseInt(parsedUserAgent.os.major) <= supportedMajorVersion
```
if `!parsedUserAgent.os.major.isNullOrEmpty()` returns true (e.g not null, not empty or has value) then it will evaluate the second part `Integer.parseInt(parsedUserAgent.os.major) <= supportedMajorVersion` with the existing value from left part

if `!parsedUserAgent.os.major.isNullOrEmpty()` returns false (e.g null, empty or no value) then it will short-circuit and won't evaluate the second part (it won't call the function  like following `Integer.parseInt("")`) which will not throw `NumberFormatException`

what do you think?

mpanne: Reasonable. I would still wrap the whole user-agent parsing in a try-catch block since this should not influence the response to the user in case of any runtime exceptions. In case of any exception the widget is just returned as usual.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -57,6 +63,18 @@ class WidgetController(
             "additionalClass" to ""
         )
 
+        try {
```
mpanne, line 66: This try catch block could be moved ("hidden") inside the `isIncompatibleOSVersion` method where the `catch` just returns `true`. This would remove the duplicate widget rendering return. 
Second option would be using `finally` instead of `catch` and removing the second return.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -127,4 +145,25 @@ class WidgetController(
     private fun setEiDClientURL(url: String): Pair<String, String> {
         return "eidClientURL" to url
     }
+
+    /*
+       Since user agent will return version in String, we need to properly convert the string to integer,
+       otherwise we cannot perform logical operations such as <= or >=
+       If major version returns empty or somehow the user agent parser cannot parse the version properly,
+       we will show to users the widget
+    */
```
mpanne, line 154: Comment can be removed

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -127,4 +145,25 @@ class WidgetController(
...
+        val incompatibleAndroidVersion = hasIncompatibleMajorVersion(parsedUserAgent, "Android", 9)
+
+        return incompatibleIOSVersion || incompatibleAndroidVersion
+    }
+
+    private fun hasIncompatibleMajorVersion(parsedUserAgent: Client, osFamily: String, supportedMajorVersion: Int): Boolean {
+        return parsedUserAgent.os.family == osFamily &&
+            !parsedUserAgent.os.major.isNullOrEmpty() &&
+            Integer.parseInt(parsedUserAgent.os.major) < supportedMajorVersion
+    }
```
mpanne, line 168: I like the refactoring!! Much clearer now.

mpanne (changes requested):

mpanne (approved):
Ready to be let off the leash imho!

code28 (approved):

PR #134, opened 2022-11-23
USEID-645: Add session id to Matomo calls

code28:


src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -36,25 +35,28 @@ class WidgetController(
     )
 
     @PostMapping("/$WIDGET_START_IDENT_BTN_CLICKED")
-    fun handleAppOpened(): ResponseEntity<String> {
+    fun handleStartIdentButtonClicked(@RequestParam(required = false) hash: String?): ResponseEntity<String> {
```
mpanne, line 38: Can we get a more descriptive name than `hash`? Maybe `sessionHash`?

src/main/resources/static/js/identifyHandler.js
```diff
@@ -1,5 +1,6 @@
+const hash = new URL(window.location).searchParams.get("hash");
```
mpanne, line 1: How about renaming this file to something like `setEIDClientURLToStartIdentButton`? Or maybe only `setEIDClientURL`?

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -96,18 +99,19 @@ class WidgetController(
...
+            widgetTracking.names.fallback,
+            hash
         )
         /*
             Documentation about the link syntax can be found in Technical Guideline TR-03124-1 – eID-Client, Part 1:
             Specifications Version 1.4 8. October 2021, Chapter 2.2 Full eID-Client
             Note: Replaced the prefix eid:// with bundesident:// to make sure only the BundesIdent app is opened
          */
-        val url = "bundesident://127.0.0.1:24727/eID-Client?${serverHttpRequest.uri.rawQuery}"
+        val url = "bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=$tcTokenURL"
```
mpanne, line 114: Nice! <3

src/main/resources/static/js/identifyHandler.js
```diff
@@ -1,5 +1,6 @@
...
+const tcTokenURLParam = window.location.hash.replace("#", "");
+
 const eidClientButton = document.getElementById("eid-client-button");
-const eidClientBaseUrl = `${
-  eidClientButton.baseURI
-}eID-Client?${window.location.hash.replace("#", "")}`;
-eidClientButton.setAttribute("href", eidClientBaseUrl);
+const eidClientURL = `${eidClientButton.baseURI}eID-Client?${tcTokenURLParam}&hash=${hash}`;
```
mpanne, line 5: Why do we add the hash to the eID-Client URL?

mpanne (approved):
Nice! I like the idea of hashing the `tcTokenUrl` to create a session id.

src/main/resources/static/js/identifyHandler.js
code28: With pleasure. 👍 Don't like the current name, either.

src/main/resources/static/js/identifyHandler.js
code28: Because we need it, when the fallback page is loaded and we want to connect that call to the session.

mpanne: 💡

PR #135, opened 2022-11-25
USEID-572: Incompatible Page — Part 2

pgurusinga:
# Problem
This improvement was raised by @phuesler and we agree that the current implementation leads to unnecessary redirect and another endpoint is another maintenance. Keep it simple.

# Proposed changes
### Old implementation:
- Any request with incompatible OS version should be redirected to `/incompatible` path

### New implementation
- Any request with incompatible OS version, the widget itself will render the `incompatible` page template.
- Path will stay same (`/widget`)
- Content Security Policy will follow as the one from widget, we don't need to set it in the `WebFilter` anymore

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyFilter.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyFilter.kt
@@ -1,6 +1,5 @@
 package de.bund.digitalservice.useid.config
 
-import de.bund.digitalservice.useid.widget.INCOMPATIBLE_PAGE
 import de.bund.digitalservice.useid.widget.WIDGET_PAGE
 import org.springframework.http.server.reactive.ServerHttpRequest
 import org.springframework.http.server.reactive.ServerHttpResponse
@@ -16,8 +15,7 @@ class ContentSecurityPolicyFilter(
 ) : WebFilter {
 
     private val widgetPagePath: PathPattern = PathPatternParser().parse("/$WIDGET_PAGE")
-    private val incompatiblePagePath: PathPattern = PathPatternParser().parse("/$INCOMPATIBLE_PAGE")
-    private val listOfPages = listOf(widgetPagePath, incompatiblePagePath)
+    private val listOfPages = listOf(widgetPagePath)
 
     override fun filter(
         serverWebExchange: ServerWebExchange,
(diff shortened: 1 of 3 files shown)


src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyFilter.kt
```diff
@@ -16,8 +15,7 @@ class ContentSecurityPolicyFilter(
 ) : WebFilter {
 
     private val widgetPagePath: PathPattern = PathPatternParser().parse("/$WIDGET_PAGE")
-    private val incompatiblePagePath: PathPattern = PathPatternParser().parse("/$INCOMPATIBLE_PAGE")
-    private val listOfPages = listOf(widgetPagePath, incompatiblePagePath)
+    private val listOfPages = listOf(widgetPagePath)
```
pgurusinga, line 18: we can keep this as List, since we want to add FALLBACK path later

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -66,7 +66,7 @@ class WidgetController(
         )
 
         if (isIncompatibleOSVersion(userAgent)) {
-            return Rendering.redirectTo("/$INCOMPATIBLE_PAGE?hostname=$hostname").build()
```
pgurusinga, line 69: `hostname` is not used anymore, but we need to keep `@RequestParam() hostname: String` in the function param, if consumer of our endpoint is requesting to `/widget` without hostname, they will get `Bad Request` response rather than OK

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -66,7 +66,7 @@ class WidgetController(
         )
 
         if (isIncompatibleOSVersion(userAgent)) {
```
mpanne, line 68: This condition should be moved to the top of the function right? So that only one tracking event is published (either "widget loaded" or "incompatible").

mpanne (approved):
Wow, I love the clear PR description! Nice! 
Only one minor comment.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
pgurusinga: we discussed this today and we will keep it as it is, since we want to see the user journey from widget loaded to incompatible page
cc: @maximilianwoltersdorf

phuesler (approved):

PR #143, opened 2022-11-30
NO-TICKET: encode tcTokenURL in fallback page

pgurusinga:
# Problem
Fallback page button does not encode `tcTokenUrl` query param, when it's passed from `/widget` page

# Proposed changes
- Add 2 test cases for encoded and not-encoded tcTokenUrl
- Add `jsoup` package for testing. Since Hamcrest does not have HTML matcher, `jsoup` parses the html string returned by the `WebTestClient` and give us more easy to understand API, such as `getElementById(String id)` for finding element with an id

### Old test
```kotlin
val body = result.responseBody?.decodeToString()
assertThat(body, containsString("class=\"container fallback\""))
```

we can have more declarative way of testing, (it could increase developer experience)

### New test
```kotlin
val responseBody: String? = result.responseBody?.decodeToString()
val parsedResponseBody = Jsoup.parse(responseBody!!)

val containerFallback = parsedResponseBody.body().firstElementChild()?.className()
val hasValidFallbackClassName = containsString("container fallback")

assertThat(containerFallback, hasValidFallbackClassName)
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -112,6 +112,7 @@ dependencies {
     testImplementation("org.testcontainers:r2dbc:1.17.3")
     testImplementation("org.testcontainers:mysql:1.17.3")
     testImplementation("org.awaitility:awaitility:4.2.0")
+    testImplementation("org.jsoup:jsoup:1.15.3")
 
     /** Spring Cloud **/
     implementation("org.springframework.cloud:spring-cloud-starter-kubernetes-client-config:2.1.3")
--- a/src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
+++ b/src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
@@ -165,26 +165,6 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
     }
 
     @Test
-    fun `fallback page should encode tcTokenURL param for identification button when the query param is not encoded`() {
-        val decodedTcTokenUrl = "https://www.foo.bar"
-        val result = webTestClient
-            .get()
-            .uri("/$FALLBACK_PAGE?tcTokenURL=$decodedTcTokenUrl")
-            .exchange()
-            .expectStatus().isOk
-            .expectBody()
-            .returnResult()
-
-        val responseBody: String? = result.responseBody?.decodeToString()
-        val parsedResponseBody = Jsoup.parse(responseBody!!)
-
-        val eidClientButton = parsedResponseBody.getElementById("eid-client-button")?.attr("href")
-        val hasCorrectUrl = containsString("bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=https%3A%2F%2Fwww.foo.bar")
-
-        assertThat(eidClientButton, hasCorrectUrl)
-    }
-
-    @Test
     fun `widget endpoint APP_OPENED should return 200`() {
         webTestClient
             .post()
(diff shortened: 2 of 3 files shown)


src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -129,9 +130,58 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
...
+
+        val eidClientButton = parsedResponseBody.getElementById("eid-client-button")?.attr("href")
+        val hasCorrectUrl = containsString("bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=https%253A%252F%252Fwww.foo.bar")
+
+        assertThat(eidClientButton, hasCorrectUrl)
+    }
+
+    @Test
+    fun `fallback page should encode tcTokenURL param for identification button when the query param is not encoded`() {
+        val decodedTcTokenUrl = "https://www.foo.bar"
```
pgurusinga, line 169: maybe this does not make any sense, since Spring decodes query param automatically

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -129,9 +130,58 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
...
+            .exchange()
+            .expectStatus().isOk
+            .expectBody()
+            .returnResult()
+
+        val responseBody: String? = result.responseBody?.decodeToString()
+        val parsedResponseBody = Jsoup.parse(responseBody!!)
+
+        val eidClientButton = parsedResponseBody.getElementById("eid-client-button")?.attr("href")
+        val hasCorrectUrl = containsString("bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=https%3A%2F%2Fwww.foo.bar")
```
pgurusinga, line 162: somehow the result is double encoded 😱 
```
✘ fallback page should encode tcTokenURL param for identification button when the query param is passed()
      java.lang.AssertionError: 
      Expected: a string containing "bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=https%3A%2F%2Fwww.foo.bar"
           but: was "bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=https%253A%252F%252Fwww.foo.bar"
```

pgurusinga: I think I know the root cause:
```kotlin
val responseBody: String? = result.responseBody?.decodeToString()
```
transform the bytearray to string and produces this on the template: `tcTokenURL&#x3D;https%253A%252F%252Fwww.foo.bar`

mpanne (approved):
Integrating a new dependency which is independent from the current framework is always a "big step" but I really like how `jsoup` makes testing / parsing HTML much more explicit. Nice!

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
mpanne: Understood and fixed, right? Can be resolved?

build.gradle.kts
```diff
@@ -112,6 +112,7 @@ dependencies {
     testImplementation("org.testcontainers:r2dbc:1.17.3")
     testImplementation("org.testcontainers:mysql:1.17.3")
     testImplementation("org.awaitility:awaitility:4.2.0")
+    testImplementation("org.jsoup:jsoup:1.15.3")
```
phuesler, line 115: What is this extension used for?

phuesler: Ah for the integration test

phuesler (approved):

build.gradle.kts
phuesler, line 115: Is it possible to declare this dependency as for testing only?

pgurusinga: as I understand `testImplementation()` adds test dependency and `implementation()` for compile and runtime...

phuesler: Gee, I think you are right. I overlooked that part.

phuesler (approved):
:rocket:

PR #145, opened 2022-12-02
USEID-583: Custom 4xx and 5xx pages

pgurusinga:
# Problem
We don't have a meaningful error page for the users and we don't have a mechanism for the user to report the error

# Proposed changes
- Add `4xx.mustache` and `5xx.mustache` inside new folder called `error` in the `template` directory. **why?** this is the way Spring knows that there are custom pages. We can do as following:
```shell
'/<templates>/error/404.<ext>'
'/<static>/error/404.html'
'/<templates>/error/4xx.<ext>'
'/<static>/error/4xx.html'
'/<templates>/error/error'
'/<static>/error/error.html'
```
For more detailed API, please check: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/reactive/error/DefaultErrorWebExceptionHandler.html

- Update YAML to store all the strings required to render the template including the email template
- Override `getErrorAttributes()` and merge default error attributes and our custom error attributes
- Use `UrlUtils.encode()` that encodes whitespace to `%20`, Java `URLEncoder.encode()` encodes whitespace to `+` sign, both are valid, but for our case, we need to use `%20`

https://user-images.githubusercontent.com/9930966/205306395-f2345cbe-cc5e-4742-a927-493662fd691f.mov


src/main/resources/templates/layout/header.mustache
```diff
@@ -8,7 +8,10 @@
     <link rel="stylesheet" href="css/sanitize.css" />
```
phuesler, line 8: We fetch CSS files individually? Is this the proposed pipeline ticket that should fix this?

phuesler (approved):
lgtm

phuesler:
What happens with the accept header is `application/json`?

pgurusinga:
@phuesler 
#### Before
```json
{
    "error": "Bad Request",
    "path": "/widget",
    "requestId": "6be0d8fc-8737",
    "status": 400,
    "timestamp": "2022-12-02T14:08:28.782+00:00"
}
````

#### After
```json
{
    "baseUrl": "http://localhost:8080",
    "error": "Bad Request",
    "errorDescription": "Bitte starten Sie die Identifizierung erneut. Sollte erneut ein Fehler auftreten, freuen wir uns über eine kurze Rückmeldung an:",
    "errorReportEmail": "hilfe@bundesident.de",
    "errorReportEmailLink": "mailto:hilfe@bundesident.de?subject=Fehlermeldung%20400&body=Liebes%20BundesIdent-Team%2C%0A%0Aich%20habe%20erneut%20eine%20Fehlermeldung%20erhalten.%0A%0AIch%20wollte%20mich%20bei%20folgendem%20Diensteanbieter%20identifizieren%3A%20%0ADer%20Fehler%20trat%20auf%2C%20nachdem%20ich%3A%20%0AIch%20benutze%20folgendes%20Ger%C3%A4t%3A%20%0ADas%20Betriebssystem%20meines%20Ger%C3%A4tes%3A%20%0AIch%20benutze%20folgenden%20Browser%3A%20%0ADie%20Version%20von%20meinem%20Browser%3A%20%0A%0AMit%20freundlichen%20Gr%C3%BC%C3%9Fen%0A",
    "errorTitle": "Es ist ein Fehler aufgetreten.",
    "metaTag": {
        "headerTitle": "Mit BundesIdent online ausweisen"
    },
    "path": "/widget",
    "requestId": "819c0969-1",
    "status": 400,
    "timestamp": "2022-12-02T14:07:10.973+00:00"
}

```

what do you think?

phuesler:
Looks good. If I understand the test and code correctly, we no longer return 5xx?

pgurusinga:
@phuesler we do still return 5xx in case there would be an error from eID server.
What I did, I try to re-use this test:
https://github.com/digitalservicebund/useid-backend-service/blob/1a46f76c7466a16945fbd2fbc5762ab3cddde777/src/test/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionControllerIntegrationTest.kt#L116-L126

which returns a 5xx status code, but I added accept header in the request (in the test above, accept JSON) => it should be HTML and spring knows already that the request is invalid that's why Spring returns `406 NOT ACCEPTABLE` instead of 500

nfelger:
> - Use `UrlUtils.encode()` that encodes whitespace to `%20`, Java `URLEncoder.encode()` encodes whitespace to `+` sign, both are valid, but for our case, we need to use `%20`.

I was going to ask why but then stumbled on the RFC. Pretty interesting!

> Software creating 'mailto' URIs likewise has to be careful to encode
> any reserved characters that are used.  HTML forms are one kind of
> software that creates 'mailto' URIs.  Current implementations encode
> a space as '+', but this creates problems because such a '+' standing
> for a space cannot be distinguished from a real '+' in a 'mailto'
> URI.  When producing 'mailto' URIs, all spaces SHOULD be encoded as
> %20, and '+' characters MAY be encoded as %2B.  Please note that '+'
> characters are frequently used as part of an email address to
> indicate a subaddress, as for example in <bill+ietf@example.org>.

https://www.rfc-editor.org/rfc/rfc6068#section-5

pgurusinga:
@nfelger thanks! I will incorporate the RFC into the comment and also I assume based on the RFC, we need also to encode the email address?

src/test/kotlin/de/bund/digitalservice/useid/statics/GlobalStaticErrorIntegrationTest.kt
```diff
@@ -0,0 +1,91 @@
+package de.bund.digitalservice.useid.statics
```
nfelger, line 1: Feels like there should also be a test for 500 error, don't you think?

nfelger (approved):
Left some smaller comments, but none should block merging

nfelger:
> @nfelger thanks! I will incorporate the RFC into the comment and also I assume based on the RFC, we need also to encode the email address?

If we were accepting user input, we should look into it, but since we know the email address that's being used, and it's unlikely to change, I guess we don't have to.

src/main/resources/application.yaml
```diff
@@ -112,3 +112,25 @@ widget:
...
+        ich habe erneut eine Fehlermeldung erhalten.
+
+        Ich wollte mich bei folgendem Diensteanbieter identifizieren: 
+        Der Fehler trat auf, nachdem ich: 
+        Ich benutze folgendes Gerät: 
+        Das Betriebssystem meines Gerätes: 
+        Ich benutze folgenden Browser: 
+        Die Version von meinem Browser: 
+
+        Mit freundlichen Grüßen
```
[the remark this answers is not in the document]
pgurusinga: yes, I agree and I already mentioned this in our daily last week, for now, we will keep the body template as it is and we could improve again

src/main/resources/templates/error/4xx.mustache
```diff
@@ -0,0 +1,17 @@
+{{>layout/header}}
+<div class="container">
+    <div class="default-error">
+        <div class="error_wrapper">
+            <img src="images/icon-error.svg" loading="lazy" alt="" class="error_icon">
```
mpanne, line 5: Suggestion for alt: "Ausrufezeichen auf rotem Grund".

src/main/resources/templates/error/4xx.mustache
```diff
@@ -0,0 +1,17 @@
...
+                <p>{{errorDescription}}
+                    <a href="{{errorReportEmailLink}}">
+                        {{errorReportEmail}}
+                    </a>
+                </p>
+            </div>
+        </div>
+    </div>
+</div>
+{{>layout/footer}}
```
mpanne, line 17: Shouldn't the `errorNotification` template be reused here since it has the same appearance?

src/main/kotlin/de/bund/digitalservice/useid/statics/GlobalStaticError.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlin.text.Charsets.UTF_8
+
+@Component
+class GlobalStaticError(
+    private val applicationProperties: ApplicationProperties,
+    private val globalStaticProperties: GlobalStaticProperties,
+    private val widgetProperties: WidgetProperties
+) : DefaultErrorAttributes() {
+
+    override fun getErrorAttributes(request: ServerRequest, options: ErrorAttributeOptions): Map<String, Any> {
```
mpanne, line 18: Cool thing you found a way to extend the JSON error response!

mpanne (approved):
Please don't forget about the `alt` 😉

src/test/kotlin/de/bund/digitalservice/useid/statics/GlobalStaticErrorIntegrationTest.kt
pgurusinga: I tried to throw error but for any request with html accept header, Spring already prevents to throw the error we defined in the controller and returns 406 instead

src/main/resources/templates/error/4xx.mustache
pgurusinga: now it's updated to use the `errorNotification`

mpanne (approved):

PR #149, opened 2022-12-07
USEID-724: Migrate templates to Thymeleaf

mpanne:
As a developer I want to use Thymeleaf templates for the Widget UI instead of Mustache to be able to use Spring means out of the box instead instead of implementing a custom solution (e.g. internationalization in templates).


src/main/resources/templates/widget.html
```diff
@@ -0,0 +1,98 @@
...
+          class="app-icon"
+          src="images/app-icon.svg"
+          th:alt="${localization.appIconImageAlt}"
+        />
+      </div>
+      <div class="headline">
+        <h1>
+          [[${localization.headlineTitleTop}]] <br />
+          [[${localization.headlineTitleBottom}]]
+        </h1>
```
pgurusinga, line 30: I just realise, as you mentioned in my commit merge, `h1` should be available once in a page, since our iframe means embedded in parent page, then we should not use `h1` right because the parent page (GST) has `h1` in place already

src/main/resources/templates/widget.html
```diff
@@ -0,0 +1,98 @@
...
+        />
+        <h2>[[${localization.identificationButton}]]</h2>
+      </a>
+      <div class="data-privacy-wrapper">
+        <a
+          th:href="${mobileUrl.dataPrivacy}"
+          target="_blank"
+          class="data-privacy-button"
+        >
+          [[${localization.dataPrivacyButton}]] ↗
```
pgurusinga, line 84: `↗` is so small and I thought it was a typo , would that make sense to change to html entity `&#8599;` doc: https://www.htmlsymbols.xyz/unicode/U+2197

pgurusinga (approved):
just side note on accessibility and html entity

src/main/resources/templates/widget.html
mpanne: Agree!

PR #154, opened 2022-12-13
USEID-725: Make use of i18n feature of spring

mpanne:
As a developer I want to use the built-in internationalisation feature of Spring to not manually maintain a mapping of text to properties class to variable in template and to be able to deliver UI in different languages to different users at runtime.


src/main/kotlin/de/bund/digitalservice/useid/statics/GlobalStaticError.kt
```diff
@@ -12,36 +13,34 @@ import kotlin.text.Charsets.UTF_8
...
         )
 
         return errorAttributes + customGlobalErrorAttributes
     }
 
     private fun createEmailReportLink(statusCode: Int): String {
-        val emailAddress = globalStaticProperties.errorView.localization.errorReportEmail
-        val errorReportSubject = globalStaticProperties.errorView.localization.errorReportSubject
-        val errorReportBody = globalStaticProperties.errorView.localization.errorReportBody
+        val emailAddress = messageSource.getMessage("error.default.report-email", null, LocaleContextHolder.getLocale())
```
pgurusinga, line 33: We could assign `LocaleContextHolder.getLocale()` to a variable:

`val currentLocale = LocaleContextHolder.getLocale()`

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -116,8 +113,9 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
         val iOSResponseBody = iosResponse.expectBody().returnResult().responseBody?.decodeToString()
         val androidResponseBody = androidResponse.expectBody().returnResult().responseBody?.decodeToString()
 
-        assertThat(iOSResponseBody, containsString(widgetProperties.errorView.incompatible.localization.headlineTitle))
-        assertThat(androidResponseBody, containsString(widgetProperties.errorView.incompatible.localization.headlineTitle))
+        val errorTitle = "Ihr Smartphone erfüllt leider nicht die Voraussetzungen der BundesIdent App."
```
pgurusinga, line 116: How does the test suite know that we are testing in `de` locale? This would break the test if we run in different locale, right? or maybe we should not test against string... 🤔

src/main/resources/messages_de.properties
```diff
@@ -0,0 +1,52 @@
...
+  \ ich habe erneut eine Fehlermeldung mit dem Status {0} erhalten. \n\
+  \n\
+  \ Ich wollte mich bei folgendem Diensteanbieter identifizieren: \n\
+  \ Der Fehler trat auf, nachdem ich: \n\
+  \ Ich benutze folgendes Gerät: \n\
+  \ Das Betriebssystem meines Gerätes: \n\
+  \ Ich benutze folgenden Browser: \n\
+  \ Die Version von meinem Browser: \n\
+  \n\
+  \ Mit freundlichen Grüßen
```
pgurusinga, line 52: I guess this is the way to write multi-line in `.properties` ? I searched the alternative for YAML, because with YAML multi-line is more expressive, requires no backward slash, but I could not find the solution 😓

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
hannesschaletzky: +1

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -139,9 +137,7 @@ class WidgetControllerIntegrationTest(@Autowired val webTestClient: WebTestClien
         assertThat(containerFallback, hasValidFallbackClassName)
 
         val errorTitle = parsedResponseBody.getElementsByClass("error_title").text()
-        val hasValidErrorTitle = containsString(widgetProperties.errorView.fallback.localization.errorTitle)
-
-        assertThat(errorTitle, hasValidErrorTitle)
+        assertThat(errorTitle, containsString("Es ist ein Fehler aufgetreten."))
```
hannesschaletzky, line 140: same here. 
How is this going to work with a different locale or do the tests always use the default locale? 
I would try and fetch the asserted value directly from the .properties file, if that is possible.

src/test/resources/application.yaml
```diff
@@ -51,65 +51,3 @@ tracking:
...
-
-        Ich wollte mich bei folgendem Diensteanbieter identifizieren: 
-        Der Fehler trat auf, nachdem ich: 
-        Ich benutze folgendes Gerät: 
-        Das Betriebssystem meines Gerätes: 
-        Ich benutze folgenden Browser: 
-        Die Version von meinem Browser: 
-
-        Mit freundlichen Grüßen
-
```
hannesschaletzky, line 115: the removal from here feels like a relief :)

hannesschaletzky (commented):
LGTM, just some comments at the tests. 
th-syntax looks nice!  

Thanks for refactoring 🙌

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
mpanne: German is the default (`message.properties`) if no locale is set. Currently the locale is resolved by checking the `Accept-Language` HTTP header (default behaviour). As long as we don't set this header for the test (or if the header includes locales which are not supported) the language will default to german which is why this test succeeds. 
In my opinion we can leave that for now and once we support multiple languages, we add tests which set the language header and validate the language of the widget accordingly. 
Other option would be to replace the hardcoded String with reading the value from the `MessageSource`.
What do you think?

src/main/resources/messages_de.properties
mpanne: Unfortunately, YAML is currently not supported for message files. I actually had the whole String in one line but `spotless` added the newlines with backslash.

src/main/kotlin/de/bund/digitalservice/useid/statics/GlobalStaticError.kt
mpanne: will do

src/main/resources/messages_de.properties
```diff
@@ -0,0 +1,52 @@
+widget.meta.title=Mit BundesIdent online ausweisen
+widget.headline-title-top=Mit BundesIdent
```
code28, line 2: I assume kebab-case is best practice here? Otherwise we use camelCase everywhere, that's why I was wondering...

src/main/resources/messages.properties
```diff
@@ -0,0 +1,52 @@
+widget.meta.title=Mit BundesIdent online ausweisen
```
code28, line 1: Is there a way to let Spring use `de` as default language without a need for duplicating this file? 🤔

code28 (approved):
Nice!

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
code28: I agree, that both ways have pros and cons, that's always a problem. What we actually wanted would be that all the used keys have a non-empty value, but that's rather another test, so let's use MessageSource here.

src/main/resources/messages.properties
mpanne: Good point. I mean we could just remove the `_de` file. This will have the desired effect that we do not have duplicate files and german is the default.

code28: That's fine for me for now! 👍

src/main/resources/messages_de.properties
mpanne: Kebab-case is recommended (and best practise afaik) for `.properties` and `.yaml` files in Spring (See [docs](https://docs.spring.io/spring-boot/docs/2.1.1.RELEASE/reference/html/boot-features-external-config.html) table 24.1). "Unfortunately" camelCase and snake_case are allowed as well and we also mix those up in our `application.yaml` but we mainly use kebab-case here. What do you mean with "we use camelCase" everywhere?

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
mpanne: See comment above.

src/main/resources/messages_de.properties
code28: I meant we use it in the code, so e.g. for variables and method names. But I see your point and let's stick with the recommendation. 👍

src/main/resources/messages.properties
mpanne: Done.

PR #162, opened 2022-12-30
NO-TICKET: Add CORS header

pgurusinga:
Widget responds the Access-Control-Allow-Origin header

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityHeadersFilter.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityHeadersFilter.kt
@@ -1,6 +1,7 @@
 package de.bund.digitalservice.useid.config
 
 import de.bund.digitalservice.useid.widget.WIDGET_PAGE
+import org.springframework.http.HttpHeaders
 import org.springframework.http.server.reactive.ServerHttpRequest
 import org.springframework.http.server.reactive.ServerHttpResponse
 import org.springframework.web.server.ServerWebExchange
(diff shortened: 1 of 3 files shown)


src/main/kotlin/de/bund/digitalservice/useid/config/SecurityHeadersFilter.kt
```diff
@@ -31,10 +32,13 @@ class ContentSecurityPolicyFilter(
         val hostNameIsAllowed = hostName?.let { contentSecurityPolicyProperties.domainIsAllowed(it) }
 
         if (hostNameIsAllowed == true) {
-            response.headers.set(
-                "Content-Security-Policy",
-                contentSecurityPolicyProperties.getCSPHeaderValue(hostName)
+            val securityHeaders = mapOf(
+                "Content-Security-Policy" to contentSecurityPolicyProperties.getCSPHeaderValue(hostName),
+                HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN to hostName,
+                HttpHeaders.VARY to HttpHeaders.ORIGIN
```
pgurusinga, line 38: Please refer to MDN:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#the_http_response_headers 

> If the server specifies a single origin (that may dynamically change based on the requesting origin as part of an allowlist) rather than the "*" wildcard, then the server should also include Origin in the [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) response header to indicate to clients that server responses will differ based on the value of the [Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) request header.

mpanne (approved):
LGTM

pgurusinga:
@mpanne could you please approve once again? you approved it when it was still in draft and I cannot merge it 😄

mpanne (approved):

PR #168, opened 2023-01-09
Make app store buttons more prominent by inverting colors

jerdesign:


code28:
Closing, because it's superceded by #180

PR #171, opened 2023-01-11
USEID-570: Add ADR for widget js-support

hannesschaletzky:
Created ADR to answer the question:

> Can the widget javascript be simplified or even removed?

[Ticket](https://digitalservicebund.atlassian.net/browse/USEID-580)

### DIFF
--- a/doc/adr/0017-mandatory-js-support.md
+++ b/doc/adr/0017-mandatory-js-support.md
@@ -1,4 +1,4 @@
-# 17. Mandatory JavaScript Support
+# 17. Integration through JavaScript files
 
 Date: 2023-01-11
 
@@ -8,9 +8,21 @@ Accepted
 
 ## Context
 
-For eServices to embed our widget logic correctly, they need to execute a JavaScript file. In terms of scaling and availability of our widget,
-we need a decision on whether we can ship the widget without having the need for enabled JavaScript on the client-side. There are ways in which we
-can provide our widget without JavaScript (e.g. NoScript), but they would require us to alter our server side logic as well as the storage of session-information like the TCTokenURL.
+All eServices should be able to integrate our service seamlessly. There are several ways for this.
+Most of the options come down to the decision between using JavaScript or not using JavaScript.
+
+Pros of having a solution without javascript:
+
+- Smaller attack surface
+- Easier to debug and test, because we could test against static html and wouldn't need a js runtime
+- Potentially support for more browsers
+
+Cons of having a solution without javascript:
+
+- We'd need to store more information on the server
+- No event tracking
+
+Our overarching goal
 
 ## Decision
 
--- a/doc/adr/0017-mandatory-js-support.md
+++ b/doc/adr/0017-mandatory-js-support.md
@@ -1,4 +1,4 @@
-# 17. Integration through JavaScript files
+# 17. Mandatory client-side JavaScript support
 
 Date: 2023-01-11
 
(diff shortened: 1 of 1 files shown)


doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Context
+
+For eServices to embed our widget logic correctly, they need to execute a JavaScript file. In terms of scaling and availability of our widget,
+we need a decision on whether we can ship the widget without having the need for enabled JavaScript on the client-side. There are ways in which we
+can provide our widget without JavaScript (e.g. NoScript), but they would require us to alter our server side logic as well as the storage of session-information like the TCTokenURL.
+
+## Decision
+
+We don't want the TCTokenURL to be stored within our server. The TCTokenURL is therefore part of the widget and no further calls to our backend are necessary in order to start the eID-flow at the eService.
```
phuesler, line 17: Is this the case in the current implementation or a consideration for future implementatoins?

hannesschaletzky: AFAIK, current understanding and kept as requirement in future. Our server should keep the least amount of data possible.

phuesler:
Pros of having a solution that doesn't require javascript:

- Smaller attack surface
- Easier to debug and test, because we could test against static html and wouldn't need a js runtime
- Potentially support for more browsers

Cons of having a solution without javascript:

- We'd need to store more information on the server
- No event tracking

code28:
> ‼️ I am unsure on the fact that with disabled JavaScript on the client side the widget will not work. Can someone help me there or explain?

Since the tcTokenUrl is only passed client-side and the tcTokenUrl is not inserted into the link without JavaScript, the widget would be displayed correctly without JavaScript, but the "Ausweisen" button would not work.

doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,47 @@
...
+  - potential security vectors
+
+Server-side approach (JavaScript disabled):
+
+- benefits:
+  - smaller attack surface
+  - easier to debug and test, because we could test against static html and wouldn't need a JavaScript runtime
+  - potentially more users because we do not rely on enabled JavaScript
+- costs:
+  - storage of unnecessary data on the server
```
pgurusinga, line 35: could we please explain in more detail what `data` means?

doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,47 @@
...
+  - difficult client-side event tracking
```
pgurusinga, line 36: what does it mean client-side event tracking is "difficult"? devs could still add event payload in the request body in the form (e.g. submitting form), but I think it's more preference.

https://developer.mozilla.org/en-US/docs/Learn/Forms/Sending_and_retrieving_form_data

doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,47 @@
...
+All eServices should be able to integrate our service seamlessly.
+ADR 0016 determines this integration by specifying the usage of a JavaScript file that creates an iFrame for displaying the widget.
+However, the widget needs to execute / fetch logic, everytime it is loaded, to work correctly.
+We need to decide whether we want to have a server-side or client-side approach to prepare the widget.
+Most of the reasons and implications come down to the decision between using client-side JavaScript or not, because a client side approach would require the user to have JavaScript enabled.
+
+Client-side approach (JavaScript enabled):
+
+- benefits:
+  - client side event tracking is possible
```
pgurusinga, line 20:
```suggestion
  - client-side event tracking is possible
```

doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,47 @@
...
+  - frugal use of data on the server (no need to store the constructed TCTokenURL on the server)
+  - additional room for client-side adjustments / use-cases
+- costs:
+  - potentially fewer users because we cannot enforce to enable JavaScript
+  - setting up a test environment requires more effort
+  - potential security vectors
+
```
pgurusinga, line 27: do you think an ad-blocker, would block the client-side approach?

doc/adr/0017-mandatory-js-support.md
```diff
@@ -0,0 +1,47 @@
...
+## Decision
+
+Our decision is driven by two major factors: We want to avoid storing unnecessary data on the server, and we want to properly track the client-side user-journey.
+Especially the improvements based on the client-side user tracking are a key feature of our product. Therefore, we choose to rely on client-side JavaScript and vote against a server-side approach.
+
+## Consequences
+
+We might face fewer user numbers and setting up a testing infrastructure will come with more effort.
+However, we will be able to reason our product and user-experience decision based on client-side data.
```
pgurusinga, line 47: We need a `<noscript>` tag to inform the user that javascript is required (I think we haven't done so)

doc/adr/0017-mandatory-js-support.md
hannesschaletzky: No, I don't think so. Since we haven't had a problem with this so far and most of us have an ad-blocker activated with our current JavaScript enabled solution.

doc/adr/0017-mandatory-js-support.md
hannesschaletzky: yes!

phuesler (approved):

pgurusinga (approved):

PR #180, opened 2023-01-16
USEID-788: Add A/B test for widget

code28:


src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -59,6 +60,14 @@ class WidgetController(
             sessionHash
         )
 
+        val abtestIsOriginal = Random.nextBoolean()
+        publishMatomoEvent(
+            widgetTracking.categories.abtesting,
+            widgetTracking.actions.abtest,
+            if (abtestIsOriginal) widgetTracking.names.abtestOriginal else widgetTracking.names.abtestVariation1,
```
pgurusinga, line 67: I am unsure `kotlin.random.Random` is the right tool for splitting the traffic. Does Kotlin Random ensure that the Boolean result is evenly distributed e.g. `50% true` and `50% false`?

If we don't split the traffic evenly, I am afraid that it might come e.g. `70% true` and `30% false`, if that is the case, then AB-Test is invalid, or am I missing something?

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetTracking.kt
```diff
@@ -7,19 +7,23 @@ class WidgetTracking {
...
+        val abtest = "MitigateFallback"
     }
 
     class Names {
         val widget = "widget"
         val startIdent = "startIdent"
         val incompatible = "unsupportedOS"
         val fallback = "fallback"
+        val abtestOriginal = "original"
+        val abtestVariation1 = "Variation1"
```
pgurusinga, line 26: I personally prefer to name the variation that reflects the change, when we change the blue button, e.g. `variation:button:blue` or rather if project-wise `variation:improve-cta-button`, so when we do multiple tests, it would be easier to identify in the UI

pgurusinga: also, should it be `variation1`, or why the first letter is uppercased?

src/main/resources/application.yaml
```diff
@@ -72,8 +72,8 @@ wellknown:
       fingerprint: "15:7E:42:A7:92:6B:A5:CA:9E:B0:29:8E:88:EE:81:0D:C3:13:E9:B5:84:41:50:28:8A:88:17:B4:14:40:FB:42"
 
 csp:
-  default-config: "default-src 'self';script-src 'self';style-src 'self';font-src 'self';img-src 'self';connect-src 'self';"
+  default-config: "default-src 'self';script-src 'self';style-src 'self' 'unsafe-inline';font-src 'self';img-src 'self';connect-src 'self';"
```
pgurusinga, line 75: why do we add `unsafe-line`?

pgurusinga (commented):
Thanks for the initiative, I raise a few questions here mostly logic about how we split the traffic evenly and naming the event

src/main/resources/application.yaml
code28: That was the quick (and a bit dirty) way to change the styling. Will add the inline styles into the css file, so that this can be removed.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetTracking.kt
code28: Well, it's not like only *one* thing changed, but a lot. That's why I considered the name ok-ish. The name "Variation1" was created in Matomo like that (also with capital first letter, don't know why...), that's why it's like this here, too.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
code28: Good point. A good (pseudo-)random generator should do exactly that. I didn't find something for kotlin random, but for `java.util.Random` it says ["Returns the next pseudorandom, uniformly distributed boolean value from this random number generator's sequence."](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html), so I assumed, the kotlin random class is at least that good.
So do I think, we should use the Java Random? Or do you have a better idea?

pgurusinga: I don't have any idea currently, but traffic split should not be our job usually, it should be the service that provides AB-Test. 🤔 

maybe we could use https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.random/as-java-random.html ?
e.g
```kotlin
import kotlin.random.Random
import kotlin.random.asJavaRandom

val abtestIsOriginal = Random.asJavaRandom().nextBoolean()
```

code28: As just discussed, we use this super simple approach for now.
About the `asJavaRandom`, the docs say that it "uses the specified Kotlin Random generator as a randomness source.", so there is no difference, so I'd keep the kotlin one.

PR #183, opened 2023-01-18
NO-TICKET: Fix and complete API docs

code28:


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -67,6 +71,9 @@ class IdentificationSessionsController(
         path = ["/{useIdSessionId}/$TCTOKEN_PATH_SUFFIX"],
         produces = [MediaType.APPLICATION_XML_VALUE]
     )
+    @Operation(summary = "Get TC token for this session")
+    @ApiResponse(responseCode = "200")
+    @ApiResponse(responseCode = "404", description = "No corresponding session found for that useIdSessionId", content = [Content()])
```
mpanne, line 76: What about adding 500 error for all endpoints? Or is that not part of the API docs since it is not intended?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -45,6 +48,7 @@ class IdentificationSessionsController(
     private val getEidInformationCallsWithErrorsCounter: Counter = Metrics.counter(METRIC_NAME_EID_SERVICE_REQUESTS, "method", "get_eid_information", "status", "500")
 
     @PostMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+    @Operation(summary = "Create session as eService")
```
mpanne, line 51: 401 and 500?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -103,7 +110,11 @@ class IdentificationSessionsController(
             )
     }
 
-    @GetMapping("/{eIdSessionId}")
+    @GetMapping("/{eIdSessionId}", produces = [MediaType.APPLICATION_JSON_VALUE])
+    @Operation(summary = "Fetch data as eService after identification was successful")
+    @ApiResponse(responseCode = "200")
+    @ApiResponse(responseCode = "404", description = "No corresponding session found for that eIdSessionId", content = [Content()])
+    @ApiResponse(responseCode = "401", description = "Authentication failed (missing or wrong api key)", content = [Content()])
```
mpanne, line 117: 500?

src/main/kotlin/de/bund/digitalservice/useid/events/EventController.kt
```diff
@@ -50,13 +53,19 @@ class EventController(eventService: EventService) {
 
     @PostMapping("/events/{widgetSessionId}/success")
     @ResponseStatus(HttpStatus.ACCEPTED)
+    @Operation(summary = "Push SSE to corresponding widget having an success value")
```
mpanne, line 56:
```suggestion
    @Operation(summary = "Push SSE to corresponding widget having a success value")
```

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
```diff
@@ -28,6 +31,9 @@ class RefreshController(private val identificationSessionService: Identification
     private val log = KotlinLogging.logger {}
 
     @GetMapping
+    @Operation(summary = "Redirect user to eService (after successful identification)")
```
mpanne, line 34: Could also happen after failed identification, couldn't it?

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
```diff
@@ -28,6 +31,9 @@ class RefreshController(private val identificationSessionService: Identification
...
+    @ApiResponse(responseCode = "303", content = [Content()])
+    @ApiResponse(responseCode = "404", description = "There is no corresponding session found for the provided eIdSessionId", content = [Content()])
```
mpanne, line 36: 500?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
mpanne, line 76: All secured endpoints should also define a `SecurityScheme`: https://www.baeldung.com/spring-openapi-global-securityscheme

src/main/kotlin/de/bund/digitalservice/useid/timebasedtokens/TimeBasedTokenController.kt
```diff
@@ -24,6 +25,7 @@ import java.util.UUID
 @ConditionalOnProperty(name = ["features.desktop-solution-enabled"], havingValue = "true")
 class TimeBasedTokenController(private val timeBasedTokenService: TimeBasedTokenService) {
     @PostMapping("/{useIdSessionId}/tokens", produces = [MediaType.APPLICATION_JSON_VALUE])
+    @Operation(summary = "Create (and return) time-based token for this session")
```
mpanne, line 28: Return type makes it clear enough? And returning the created object is best practice in REST, isn't it? 
This comment is really useless. It's so unimportant but now I already wrote it. 
```suggestion
    @Operation(summary = "Create time-based token for this session")
```

mpanne (changes requested):
Great initiative! Already a big improvement.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
code28: Yeah, was wondering about that as well... I thought, that 500 can always happen, when there is some unexpected error, so I didn't document that explicitly. The other response codes have a special semantics for each request, so it's more helpful.
What do you think? Adding 500 everywhere?

mpanne (approved):
None of the extra work is mandatory I realised. Spent as much time as you think would be worth it.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
mpanne: Just [a random response in a random thread](https://stackoverflow.com/questions/54238746/should-internal-error-500-be-part-of-swagger-api-documentation) but I kind of agree. So let's **not** document 500.

mpanne: Sorry for commenting it everywhere then.

code28: No, it's good to have a second thought about it 👍

PR #186, opened 2023-01-23
USEID-788: Remove A/B test and switch to widget alternative

mpanne:
The A/B test results showed that there are less visits of the fallback page with the alternative design of the widget. We switch to the new design now and remove the A/B test.


pgurusinga (approved):
LGTM

PR #187, opened 2023-01-23
UseID 791: Add device information to tracking

hannesschaletzky:
**Goal**
We have a certain number of issues, where tracking the device information would clearly help to better understand limitations of our implementation. Hence, we need to add device information to our widget tracking.

**PR Content**
We are tracking the useragent in matomo now. More fields could be added, but need customisation then. These are the default tracking properties used now and given to Matomo with the "ua" query parameter field. 
Viewable in Matomo UseidPlayground (SiteID: 5)

image

image


src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -35,32 +36,34 @@ class WidgetController(
     )
 
     @PostMapping("/$WIDGET_START_IDENT_BTN_CLICKED")
-    fun handleStartIdentButtonClicked(@RequestParam(required = false, name = "hash") sessionHash: String?): ResponseEntity<String> {
+    fun handleStartIdentButtonClicked(@RequestParam(required = false, name = "hash") sessionHash: String?, @RequestHeader(HttpHeaders.USER_AGENT) userAgent: String): ResponseEntity<String> {
```
mpanne, line 39: A `@RequestHeader` is required by default (if you do not explicitly set `required=false`. Either make this parameter optional, or take it for granted that it is never null since you use `String?` when handing the parameter to the `publishMatomoEvent` method. 
Since it is completely up to the caller (i.e. user) if the `user-agent` is set, I would suggest to make the parameter optional since otherwise the caller will get a `400 Bad Request` response.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -77,12 +80,13 @@ class WidgetController(
     }
 
     @GetMapping("/$FALLBACK_PAGE")
-    fun getUniversalLinkFallbackPage(model: Model, @RequestParam tcTokenURL: String, @RequestParam(required = false, name = "hash") sessionHash: String?): Rendering {
+    fun getUniversalLinkFallbackPage(model: Model, @RequestParam tcTokenURL: String, @RequestParam(required = false, name = "hash") sessionHash: String?, @RequestHeader(HttpHeaders.USER_AGENT) userAgent: String): Rendering {
```
mpanne, line 83: Same here

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -104,8 +108,8 @@ class WidgetController(
             .build()
     }
 
-    private fun publishMatomoEvent(category: String, action: String, name: String, sessionId: String?) {
-        val matomoEvent = MatomoEvent(this, category, action, name, sessionId)
+    private fun publishMatomoEvent(category: String, action: String, name: String, sessionId: String?, userAgent: String?) {
```
mpanne, line 111: See comment above. Either make the `userAgent` optional or use the nullsafe type here.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -131,12 +135,13 @@ class WidgetController(
             Integer.parseInt(parsedUserAgent.os.major) < supportedMajorVersion
     }
 
-    private fun handleRequestWithIncompatibleOSVersion(sessionHash: String?): Rendering {
+    private fun handleRequestWithIncompatibleOSVersion(sessionHash: String?, @RequestHeader(HttpHeaders.USER_AGENT) userAgent: String): Rendering {
```
mpanne, line 138:
```suggestion
    private fun handleRequestWithIncompatibleOSVersion(sessionHash: String?, userAgent: String): Rendering {
```

mpanne (changes requested):
Looks good! Only having the header as a required parameter could lead to problems. Maybe worth testing it?

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
hannesschaletzky: Good catch! I marked it as optional by using String?. 
Tested it and routes work when not sending any user agent.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -35,32 +36,34 @@ class WidgetController(
     )
 
     @PostMapping("/$WIDGET_START_IDENT_BTN_CLICKED")
-    fun handleStartIdentButtonClicked(@RequestParam(required = false, name = "hash") sessionHash: String?): ResponseEntity<String> {
+    fun handleStartIdentButtonClicked(@RequestParam(required = false, name = "hash") sessionHash: String?, @RequestHeader(HttpHeaders.USER_AGENT, required = false) userAgent: String?): ResponseEntity<String> {
```
pgurusinga, line 39: shouldn't we use a named argument? or the first argument is for the `name` ?

```suggestion
    fun handleStartIdentButtonClicked(@RequestParam(required = false, name = "hash") sessionHash: String?, @RequestHeader(name = HttpHeaders.USER_AGENT, required = false) userAgent: String?): ResponseEntity<String> {
```

hannesschaletzky: it works without it as well, but I will add the name.

mpanne (approved):
LGTM

PR #189, opened 2023-02-01
NO-TICKET: Update app properties for sentry

pgurusinga:
**As** a developer,
**I want** to check whether the error will be sent to Sentry or not
**so that** I can differentiate the production error and staging error
**and** I don't pollute production error events

-- 

if we don't set the environment explicitly, Sentry assumes the production environment.

### DIFF
--- a/src/main/resources/application-local.yaml
+++ b/src/main/resources/application-local.yaml
@@ -48,6 +48,9 @@ csp:
 features:
   desktop-solution-enabled: true
 
-sentry:
-  environment: staging
-  traces-sample-rate: 1.0
+# Uncomment this section if you would like to validate that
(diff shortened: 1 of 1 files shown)


phuesler:
Do we want to sent exceptions to Sentry in development?

phuesler:
I argue that we would set this for our staging deploy. In development, unless we are testing Sentry, we don't need exceptions sent to it.

pgurusinga:
@phuesler true, it's just sometimes if it's not written, people tend to forget. I'd close this PR then.

phuesler:
Is there a way to have it commented out?

pgurusinga:
@phuesler yup, it can be done.

PR #200, opened 2023-02-17
NO-TICKET: Add Github actions for manual deploy to staging

JulianRoesner:
This would add two workflows that we could run manually. One to create an image from any branch or commit and one to push any image to staging. This would allow us to deploy a testing branch or an older image quite easily to staging.

I've decided against keeping any other jobs in this action because a) the tests and analysis should have been performed on the branch already and b) we might sometimes want to deploy something to staging for testing purposes where some of the other stages might fail (e.g. for an adapted image for load-testing).

# Open questions
- What do you think about this approach?
- Is there something that I overlooked
- Do you see any errors or problems in the workflow files?


pgurusinga:
Thanks for the initiative.

- How can I trigger this action to run and manually deploy to staging on the PR created? a readme update would be great
- Is it possible if we could attach a label to a Pull Request, let's say the label `STAGING_DEPLOYMENT` then it will trigger the deployment.
- I have a thought that we could just:
	- build and publish the image locally from our machine
	- get the image tag
	- change the infrastructure

but I see that we, as humans, make errors and we might change the production image tag instead of staging

JulianRoesner:
Thanks for looking at the PR and your comments.

> How can I trigger this action to run and manually deploy to staging on the PR created? a readme update would be great

Good point, I have added a section to the Readme. After merging they will appear as specific workflows [here](https://github.com/digitalservicebund/useid-backend-service/actions/) and you can then manually run the workflow there. You can have a look at the ones in erica [here](https://github.com/digitalservicebund/erica/actions/workflows/build-image.yml). 

> Is it possible if we could attach a label to a Pull Request, let's say the label STAGING_DEPLOYMENT then it will trigger the deployment.

Yes, I would assume that is possible. However, imo there are two downsides to that:
- You always have to create a PR first. Sometimes you just want to quickly test something or even quickly use an old image to recreate some bug and creating a PR just for this purpose is a lot more work. For me, having the manual workflow is open to more general usages.
- We do not want anyone outside of the team to trigger a deployment. GitHub Actions are not run in PRs from outside collaborators. However, we also need to make sure that adding the label to PRs from inside collaborators is not possible and so on. 

Wdyt?

>I have a thought that we could just:
build and publish the image locally from our machine
get the image tag
change the infrastructure
but I see that we, as humans, make errors and we might change the production image tag instead of staging

Yes, this should always be possible. However, I agree with you that it's error prone. Furthermore, everyone needs to know how to do it and this way it's at least a little bit hidden behind some abstraction (and e.g. also tracked as a deployment to staging and the gradle wrapper is validated).

JulianRoesner:
> sorry for another comment, I just learned that we can reuse our workflow. So we could include those files you created as a partial main one for the main pipeline build, right?

Uh, I did not know that. So, basically we could use the workflow like an action? That sounds great.

pgurusinga:
> > sorry for another comment, I just learned that we can reuse our workflow. So we could include those files you created as a partial main one for the main pipeline build, right?
> 
> Uh, I did not know that. So, basically we could use the workflow like an action? That sounds great.

yes, I am just afraid if we create another partial duplication of the workflow file, we might forget to update it in the future in case we changed something in the main workflow file. but I guess we can merge this now and improve

JulianRoesner:
> yes, I am just afraid if we create another partial duplication of the workflow file, we might forget to update it in the future in case we changed something in the main workflow file. but I guess we can merge this now and improve later

Yes, I think that's a valid concern and I think we should do it before the merge because otherwise, we will probably keep it that way. If you have time, feel free to change it. Otherwise, I will try to have a look. 
Thanks for thinking that through 👍

pgurusinga:
I also can contribute when I have time after finishing the GST stories

mpanne:
I actually really like your idea @pgurusinga to extract the reusable parts to not have them duplicated.

pgurusinga:
@mpanne feel free to improve, I don't have the time at the moment 😢

JulianRoesner:
I've added the improvement. @pgurusinga do you want to have another look?

pgurusinga (approved):
Looks amazing.

PR #202, opened 2023-02-20
Useid 727 and Useid 696 add tenant id to prometheus metrics and event calls

phuesler:
This PR adds the tenant id, essentially the customer id, to multiple metrics in order to be able to distinguish between tenants operationally. Questions like: How much traffic from tenant A? Does this error only happen with tenant B?

How do we know the tenant id?

* Configuration passed in during deployment containing the mappings for both api keys and CSP header configuration
* Mapped from either the API key or a host query param
* useIdSessionId if we store the tenant id on session creation
* Passed in as a parameter
* Get session id from tcTokenUrl and retrieve the session

Where do we need this information?

* Spring boots generic controller http metrics
* The controllers to pass the tenant id to
  * Custom eid server metrics
  * matomo events

While the cases for the api routes and the widget page are straightforward to do in a request filter assigning a request attribute, other routes are not. We can store the tenant id in the session, to make it available to all the routes that retrieve the session. That still leaves a few routes where we don't have the tenant id.

| Route  | Source of tenant id | Spring HTTP metric | Matomo Event | Custom metric |  
| ------------- | ------------- | ------------- | ------------- | ------------- |
| /api/v1/events/{widgetSessionId} | ?  | no  | no   | no  |
| /api/v1/events/{widgetSessionId}/error  | ?  | no  | no  | no  |
| /api/v1/events/{widgetSessionId}/success  | ?  l  | no  | no  | no  |
| /api/v1/identification/sessions  | api key | yes  | no  | no  |
| /api/v1/identification/sessions/{eIdSessionId} | api key  | yes  | no  | counter.get_eid_information  |
| /api/v1/identification/sessions/{useIdSessionId}/tc-token | useIdSessionId or RefreshAddress  | possible?  | no  | counter.get_tc_token  |
| /api/v1/identification/sessions/{useIdSessionId}/tokens | useIdSessionId  | Possible?  |  no  | no  |
| /api/v1/identification/sessions/{useIdSessionId}/tokens/{tokenId}  | useIdSessionId  | Possible?  |  no  | no  |
| /api/v1/identification/sessions/{useIdSessionId}/transaction-info | useIdSessionId  | Possible?  | no  | no  |
| /refresh | eidSessionId  | Possible?  | no | no  |
| /eid-Client  | tcTokenUrl?  | Not possible?  | loaded.fallback  | no  |
| /startIdentButtonClicked | sent by js?  | yes | butttonPressed.startIdent  | no  |
| /widget  | hostname query param  | yes  | loaded.widget, loaded.incompatible  | no  |
| /qrcode-widget l  | hostname query param  | yes  | no | no  |
| /  |  ?  | no  | no  | no  |
| /.well-known/apple-app-site-association  | ?  | no  | no  | no  |
| /.well-known/assetlinks.json | no  | no  | no  | no  |

## Refactoring of the code for easier tenant handling

This PR unifies api key, csp and tenant id in one configuration object. Everything is now attached to the tenant configuration, That changes the code for authentication handling as well.

## Which metrics do we really need to have with a tenant id and which are optional? 

Getting everything in requires a lot of code changes throughout the stack. Is the squeeze worth the juice here? For the generic spring http metrics, we can do away without the tenant id for the time being.

* matomo events (loaded.widget, loaded.incompatible, buttonPressed.startIdent)
* Custom counters (get_id_information, get_tc_token)
* spring http metrics for api authenticated routes

- [x] Matomo event loaded.widget, loaded.incompatible
- [x] Matomo event buttonPressed.startIdent
- [x] Matomo event fallback page
- [x] counter.get_id_information
- [x] counter.get_tc_token
- [x] spring http metrics for api authenticated calls

## Matomo integration

Use a custom dimension to track the tenant id. The documentation recommends it over custom variables. We assume it is the dimension with index 1. See the [docs](https://matomo.org/faq/reporting-tools/create-track-and-manage-custom-dimensions/) for more information. We still need to set things up on Matomo side.

## Open Issues

- [x] getTenantId returns Rubbish 
- [x] The string "unknown" for an uknown tenant id is spread all over the code base
- [x] Make sure we check against a valid list of tenant ids
- [x] No explicit testing
- [x] Non idiomatic kotlin
- [x] Fix all tests


phuesler:
An optimisation would be memoization of the counters. This way we would create fewer objects, which means less garbage collection. The impact, however, would be fairly small.

phuesler:
Changing the configuration of the csp properties might break existing deployments because we'll be moving from a list with values to a list with objects.

phuesler:
Help wih idiomatic Kotlin needed. The null checks all over the place work against Kotlin.

phuesler:
One issue with the code: The tenant_id "unknown" is sprinkled all over the code. Using a constant or wrapping it inside a class might be a way out. Of course, a class will increase coupling.

phuesler:
Issue: tenant_id is used all over the place. How to best encapsulate this cross cutting concern without introducing a tight coupling to let's say, a TenantIdBuilder class?

mpanne (commented):
I really like the newly introduced `tenant`. I just added my thoughts as comments. Let's talk about it.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -129,9 +135,9 @@ class IdentificationSessionsController(
         try {
             val eidService = EidService(eidServiceConfig)
             userData = eidService.getEidInformation(eIdSessionId.toString())
-            getEidInformationCallsSuccessfulCounter.increment()
+            createCounter("get_eid_information", "200", tenantId).increment()
         } catch (e: Exception) {
-            getEidInformationCallsWithErrorsCounter.increment()
+            createCounter("get_eid_information", "500", tenantId).increment()
```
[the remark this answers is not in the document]
phuesler: It needs to be ensured that we create a new counter per distinct tenant id. That's how I remember the metrics work.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -128,15 +129,21 @@ class WidgetController(
...
             sessionHash,
             userAgent,
+            tenantId,
         )
         val modelAndView = ModelAndView(INCOMPATIBLE_PAGE)
         modelAndView.addAllObjects(defaultViewHeaderConfig)
         return modelAndView
     }
+
+    // FIXME: Check against a valid list of tenant ids
```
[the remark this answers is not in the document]
phuesler: With the Tenant model, we can do this now

src/main/kotlin/de/bund/digitalservice/useid/config/ContentSecurityPolicyProperties.kt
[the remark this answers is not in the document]
hannesschaletzky: Good idea! 
Done ✅

src/main/resources/application-local.yaml
```diff
@@ -1,14 +1,22 @@
 app:
   base-url: http://localhost:8080
 
-api:
-  apiKeys:
-    - keyvalue: foobar
+tenant:
```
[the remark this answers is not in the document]
hannesschaletzky: added this as a todo in my comment.

src/main/resources/templates/widget.html
```diff
@@ -82,14 +82,38 @@ <h2 th:text="#{widget.identification-button}"></h2>
...
-      th:if="${isWidget}"
-      type="text/javascript"
-      src="js/setEIDClientURL.js"
-    ></script>
+    <!--/*
+      -  Wrap everything inside closures to not add to the global namespace
+      -  No document ready needed because we are at the end of the document
+      -  Inlining javascript to save additional requests to the server
+      -  Render tenantId as a javascript variable
+      -  Comment won't be visible in the html according to Thymeleaf docs
```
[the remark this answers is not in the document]
hannesschaletzky: I don't know what that means actually

hannesschaletzky:
Open things: 

- [ ] Think about a better name than tenant.tenants in app.yaml
- [ ] "Inlining javascript to save additional requests to the server" in widget.html, according to [this comment](https://github.com/digitalservicebund/useid-backend-service/pull/202#discussion_r1145974121)
- [ ] clarify in which cases an "unkown" tenant in the tenantFilter.kt can be reached 
- [ ] extensive testing!

src/test/resources/application.yaml
```diff
@@ -1,14 +1,18 @@
...
+    - id: integration_test_1
+      apiKey: valid-api-key-1
+      refreshAddress: valid-refresh-address-1
+      dataGroups: [ DG1,DG2 ]
+      allowedHosts: ["i.am.allowed.1", "i.am.also.allowed.1"]
+    - id: integration_test_2
+      apiKey: valid-api-key-2
+      refreshAddress: valid-refresh-address-2
+      dataGroups: [ DG4,DG17 ]
+      allowedHosts: ["i.am.allowed.2", "i.am.also.allowed.2"]
```
pgurusinga, line 15: @hannesschaletzky

I am a bit confused with `allowedHosts`. it has a data type of array, which means we allow one or more hosts with the same API key to access our backend (including widgets embedded in multiple hosts). is that a product decision? is there any security risk?

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -31,7 +31,9 @@ class MatomoTrackingService(trackingProperties: TrackingProperties, private val
     fun constructEventURL(e: MatomoEvent): String {
         val session = e.sessionId?.let { "&uid=$it" } ?: ""
         val userAgent = e.userAgent?.let { "&ua=${UriUtils.encode(e.userAgent, UTF_8)}" } ?: ""
-        val url = "$domain?idsite=$siteId&rec=1&ca=1&e_c=${e.category}&e_a=${e.action}&e_n=${e.name}$session$userAgent"
+        // custom dimension for tenantId -> https://matomo.org/faq/reporting-tools/create-track-and-manage-custom-dimensions/
+        val tenantId = e.tenantId?.let { "&dimension3=${UriUtils.encode(e.tenantId, UTF_8)}" } ?: ""
```
maximilianwoltersdorf, line 35: I think dimension3 must be dynamic, on matomo production (site **useid** it is dimension1)

hannesschaletzky: Thanks for pointing that out 👍
I will think about a good solution!

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -117,21 +115,21 @@ class IdentificationSessionsController(
...
-        if (apiKeyDetails.refreshAddress != identificationSession.refreshAddress) {
+        if (tenant.refreshAddress != identificationSession.refreshAddress) {
             log.error("API key differs from the API key used to start the identification session.")
             return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build()
         }
         try {
             val eidService = EidService(eidServiceConfig)
             userData = eidService.getEidInformation(eIdSessionId.toString())
-            getEidInformationCallsSuccessfulCounter.increment()
+            metricsService.incrementCounter(METRIC_NAME_EID_INFORMATION, "200", tenant.id)
```
mpanne, line 130: Yes to this.

src/main/resources/static/js/initWidget.js
mpanne, line 1: This is never used, or is it?

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -65,7 +85,24 @@ class WidgetControllerIntegrationTest(
             .expectHeader()
             .valueEquals(
                 "Content-Security-Policy",
-                "some default value;frame-ancestors 'self';",
+                contentSecurityPolicy.getDefaultCSPHeaderValue(),
```
mpanne, line 88: If you use the methods of the `contentSecurityPolicy` component to validate the result, you are not testing those methods any longer since the production code uses the same. So if we mess up those, the tests will not break. Either add new tests for the methods of `contentSecurityPolicy` or restore the behaviour from before by comparing to the expected outcome (String values).

src/main/kotlin/de/bund/digitalservice/useid/config/CustomWebClientExchangeTagsProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+import de.bund.digitalservice.useid.tenant.REQUEST_ATTR_TENANT
+import de.bund.digitalservice.useid.tenant.Tenant
+import io.micrometer.core.instrument.Tag
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.boot.actuate.metrics.web.servlet.WebMvcTagsContributor
+import org.springframework.stereotype.Component
+
+@Component
+class CustomWebClientExchangeTagsProvider : WebMvcTagsContributor {
```
mpanne, line 13: Could we add a comment here what this class does?

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityHeadersFilter.kt
```diff
@@ -24,15 +28,17 @@ class SecurityHeadersFilter(
 
         if (!pathIsValidWidgetPages) return filterChain.doFilter(request, response)
 
-        val hostName: String? = request.getParameter("hostname")
-        val hostNameIsAllowed = hostName?.let { contentSecurityPolicyProperties.domainIsAllowed(it) }
+        val host = request.getParameter("hostname") ?: ""
+        val tenant = tenantProperties.findByAllowedHost(host)
```
mpanne, line 32: Why do the resolving again which is already done by the `ResolveTenantFilter?`. Make sure this filter runs after the resolving has happened and read the tenant from the request.

src/main/kotlin/de/bund/digitalservice/useid/tenant/ResolveTenantFilter.kt
```diff
@@ -0,0 +1,35 @@
...
+                val host = request.getParameter("hostname") ?: ""
+                tenantProperties.findByAllowedHost(host) ?: fallbackTenant
+            } else if (request.getParameter(PARAM_NAME_TENANT_ID) != null) {
+                val id = request.getParameter(PARAM_NAME_TENANT_ID)
+                tenantProperties.findByTenantId(id) ?: fallbackTenant
+            } else {
+                fallbackTenant
+            }
+
+        // assign entire object, because the WidgetController needs access to multiple properties instead of only the id
```
mpanne, line 31: This logic might change in the future. I personally would remove this comment. The code would break on compile time if someone changes it, so no need for a warning imho.

src/main/kotlin/de/bund/digitalservice/useid/tenant/Tenant.kt
```diff
@@ -0,0 +1,10 @@
+package de.bund.digitalservice.useid.tenant
+
+class Tenant {
+    lateinit var id: String
+    lateinit var apiKey: String
+    lateinit var refreshAddress: String
+    var dataGroups: List<String> = emptyList()
+    var allowedHosts: List<String> = emptyList()
+    var cspNonce: String = ""
```
mpanne, line 9: Why default empty String here and not `lateinit` as well?

src/main/kotlin/de/bund/digitalservice/useid/tenant/TenantProperties.kt
mpanne, line 1: Tests for that?

mpanne (approved):
Good work! Just some comments regarding cleanup or little improvements. 
Also, please do not forget that we need to adapt all API key env variables in our useid-infra repository before merging and deploying. To allow a graceful rolling update, we should first add the new tenant config and remove the old api key config after successful deployment.

src/main/resources/static/js/initWidget.js
hannesschaletzky: I don't know why this is committed, I removed it locally. 
Weird.

src/test/kotlin/de/bund/digitalservice/useid/refresh/RefreshControllerIntegrationTest.kt
```diff
@@ -61,9 +61,9 @@ class RefreshControllerIntegrationTest(@Autowired val webTestClient: WebTestClie
             .expectStatus()
             .is3xxRedirection
             .expectHeader()
-            // "some-refresh-address" value is defined in the test application.yaml
+            // "valid-refresh-address" value is defined in application.yaml under test
```
code28, line 64: Just a comment, but still..
```suggestion
            // "valid-refresh-address-1" value is defined in application.yaml under test
```

code28 (approved):
Nice!

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -117,21 +115,21 @@ class IdentificationSessionsController(
...
         authentication: Authentication,
     ): ResponseEntity<GetResultResponseType> {
-        val apiKeyDetails = authentication.details as ApiKeyDetails
+        val tenant = authentication.details as Tenant
 
         val userData: GetResultResponseType?
         val identificationSession = identificationSessionService.findByEIdSessionId(eIdSessionId)
             ?: return ResponseEntity.status(HttpStatus.NOT_FOUND).build()
-        if (apiKeyDetails.refreshAddress != identificationSession.refreshAddress) {
+        if (tenant.refreshAddress != identificationSession.refreshAddress) {
```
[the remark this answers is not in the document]
mpanne: YES! PLEASE! Good catch!

src/main/kotlin/de/bund/digitalservice/useid/tenant/Tenant.kt
hannesschaletzky: Because this is not being injected by the application.yaml, thus a default value is better because it will run into problems when tried to be accessed. 
My assumption.

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
hannesschaletzky: You are totally right. Thanks for clarifying on this one 🙌

src/main/kotlin/de/bund/digitalservice/useid/tenant/Tenant.kt
mpanne: But don't we want to run into problems if we try to access it but it is not set? Otherwise the nonce would have no value and we would not even notice it. Only the user would since the scripts would not be executed.

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -31,7 +31,9 @@ class MatomoTrackingService(trackingProperties: TrackingProperties, private val
...
+        val url = "$domain?idsite=$siteId&rec=1&ca=1&e_c=${e.category}&e_a=${e.action}&e_n=${e.name}$session$userAgent$tenantId"
```
mpanne, line 36: Another idea that just came to my mind. It might make sense to use a [UriComponentsBuilder](https://www.baeldung.com/spring-uricomponentsbuilder) here. The URI is getting more complex and this would make it more resilient and may add some readability.

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -34,14 +36,18 @@ class SecurityConfig(
...
+                TenantAuthenticationFilter(authenticationManager),
                 AnonymousAuthenticationFilter::class.java,
             )
+            .addFilterAfter(
+                ResolveTenantFilter(tenantProperties),
+                AuthorizationFilter::class.java, // Last filter in the Spring Security filter chain
+            )
+            .addFilterAfter(
+                SecurityHeadersFilter(),
+                AuthorizationFilter::class.java, // Last filter in the Spring Security filter chain
```
mpanne, line 49: To make sure the filter are added in the right order, we should use the chaining feature. So let's add the `SecurityHeaderFilter` after the `ResolveTenantFilter`.
```suggestion
                ResolveTenantFilter::class.java,
```

src/main/kotlin/de/bund/digitalservice/useid/config/CustomWebClientExchangeTagsProvider.kt
hannesschaletzky: This was Patricks work. I added a comment with my knowledge. Feel free to edit and expand @phuesler

hannesschaletzky:
Things to do in infra repo:
- [ ] remove "https" from tracking url
- [ ] create dimension values in stag and prod (tacking.matomo.dimensionIdTenant)
- [ ] create tenants based on apikeyConfig in stag and prod
- [ ] if everything works, remove apikeyConfig

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -20,24 +20,39 @@ import kotlin.text.Charsets.UTF_8
...
-        val url = "$domain?idsite=$siteId&rec=1&ca=1&e_c=${e.category}&e_a=${e.action}&e_n=${e.name}$session$userAgent$tenantId"
-        log.debug { url }
-        return url
+        val tenantId = e.tenantId?.let { "&dimension$dimensionIdTenant=${encode(e.tenantId)}" } ?: ""
+
+        return UriComponentsBuilder
+            .newInstance()
+            .scheme("https")
+            .host(domain)
+            .path("?idsite=$siteId")
```
code28, line 41: Das praktische am `UriComponentsBuilder` ist gerade, dass wir uns nicht selbst Gedanken darum machen müssen, ob `?` oder `&`, dafür gibt es `.queryParam()`. Und für optionale sollte `.queryParamIfPresent()` auch helfen, sodass wir das gar nicht mehr selbst machen müssen oben.

phuesler:
This has come a long way. Anything left I can help with?

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
hannesschaletzky: UriComponentBuilder encoded query parameter nicht. Es ist also für unsere Zwecke nicht richtig einsetzbar. 
See for example [here](https://stackoverflow.com/questions/18138011/url-encoding-using-the-new-spring-uricomponentsbuilder)
Ich versuche trotzdem was gutes zu bauen

hannesschaletzky:
@phuesler You could have a final look today so I could merge beginning of next week.

src/main/kotlin/de/bund/digitalservice/useid/metrics/MetricsService.kt
```diff
@@ -0,0 +1,30 @@
+package de.bund.digitalservice.useid.metrics
+
+import de.bund.digitalservice.useid.tenant.PARAM_NAME_TENANT_ID
+import io.micrometer.core.instrument.Counter
+import io.micrometer.core.instrument.Metrics
+import org.springframework.stereotype.Service
+
+@Service
+class MetricsService {
```
phuesler, line 9: An improvement is to keep a record of counters per name and tag list. This way, we would create fewer counters objects per request.

src/main/kotlin/de/bund/digitalservice/useid/metrics/MetricsService.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    fun incrementSuccessCounter(method: String, tenantId: String) {
+        createCounter(method, "200", tenantId).increment()
+    }
+
+    fun incrementErrorCounter(method: String, tenantId: String) {
+        createCounter(method, "500", tenantId).increment()
+    }
+
+    private fun createCounter(method: String, status: String, tenantId: String): Counter {
```
phuesler, line 19: Here we could check a list of existing counters with the key method,tenant_id and reuse that instead of creating a new one. Unless the Metrics.counter method already does that.

src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingProperties.kt
```diff
@@ -3,11 +3,9 @@ package de.bund.digitalservice.useid.tracking
 import jakarta.validation.Valid
 import jakarta.validation.constraints.NotBlank
 import org.springframework.boot.context.properties.ConfigurationProperties
-import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Component
 import org.springframework.validation.annotation.Validated
 
-@Profile("!local")
```
phuesler, line 10: Do we need to keep this so that the tests run?

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -20,26 +21,43 @@ import kotlin.text.Charsets.UTF_8
...
+    private val dimensionIdTenant = trackingProperties.matomo.dimensionIdTenant
 
-    fun constructEventURL(e: MatomoEvent): String {
-        val session = e.sessionId?.let { "&uid=$it" } ?: ""
-        val userAgent = e.userAgent?.let { "&ua=${UriUtils.encode(e.userAgent, UTF_8)}" } ?: ""
-        val url = "$domain?idsite=$siteId&rec=1&ca=1&e_c=${e.category}&e_a=${e.action}&e_n=${e.name}$session$userAgent"
-        log.debug { url }
-        return url
+    fun constructEventUrl(e: MatomoEvent): String {
+        return UriComponentsBuilder
```
phuesler, line 33: Nice refactor

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -20,26 +21,43 @@ import kotlin.text.Charsets.UTF_8
...
+            .queryParamIfPresent("uid", enc(e.sessionId))
+            .queryParamIfPresent("ua", enc(e.userAgent))
+            // custom dimension for tenantId -> https://matomo.org/faq/reporting-tools/create-track-and-manage-custom-dimensions/
+            .queryParamIfPresent("dimension$dimensionIdTenant", enc(e.tenantId))
+            .build()
+            .toString()
+    }
+
+    // UriComponentBuilder does not support query param encoding
+    private fun enc(input: String?): Optional<String> {
```
phuesler, line 52:
```suggestion
    private fun uriEncode(input: String?): Optional<String> {
```

Idiomatic change. `enc` imo, is short for encryption, not encoding.

src/main/kotlin/de/bund/digitalservice/useid/widget/WidgetController.kt
```diff
@@ -76,40 +87,48 @@ class WidgetController(
...
         /*
             Documentation about the link syntax can be found in Technical Guideline TR-03124-1 – eID-Client, Part 1:
             Specifications Version 1.4 8. October 2021, Chapter 2.2 Full eID-Client
             Note: Replaced the prefix eid:// with bundesident:// to make sure only the BundesIdent app is opened
          */
-        val url = "bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=${URLEncoder.encode(tcTokenURL, UTF_8)}"
-
         val widgetViewFallbackConfig = mapOf(
-            setEiDClientURL(url),
+            "eidClientURL" to "bundesident://127.0.0.1:24727/eID-Client?tcTokenURL=${URLEncoder.encode(tcTokenURL, UTF_8)}",
```
phuesler, line 110: To be consistent with the rest of the code base, we'd have to move the base url out to a constant.

src/main/resources/templates/widget.html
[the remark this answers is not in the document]
phuesler: What what means? i'm happy to elaborate

src/test/kotlin/de/bund/digitalservice/useid/widget/WidgetControllerIntegrationTest.kt
```diff
@@ -33,39 +49,63 @@ class WidgetControllerIntegrationTest(
...
-                "foo.bar",
+                allowedHost,
             )
             .expectHeader()
             .valueEquals(HttpHeaders.VARY, HttpHeaders.ORIGIN)
     }
 
     @Test
-    fun `widget endpoint returns default Content-Security-Policy when the request URL is invalid`() {
+    fun `widget endpoint returns default Content-Security-Policy when query parameter hostname has forbidden value`() {
```
phuesler, line 81: good to test this

phuesler:
@hannesschaletzky I went through the code and added the comments. I also updated the PR details with information on the new tenant objects.

I think this extension and refactoring of the code makes it better. Good work.

src/main/kotlin/de/bund/digitalservice/useid/tenant/ResolveTenantFilter.kt
```diff
@@ -0,0 +1,34 @@
...
+            if (request.servletPath.equals(REQUEST_PATH_WIDGET)) {
+                val host = request.getParameter("hostname") ?: ""
+                (tenantProperties.findByAllowedHost(host) ?: WidgetDefaultTenant()).apply {
+                    cspHost = host
+                }
+            } else if (request.getParameter(PARAM_NAME_TENANT_ID) != null) {
+                val id = request.getParameter(PARAM_NAME_TENANT_ID)
+                tenantProperties.findByTenantId(id) ?: FallbackTenant()
+            } else {
+                FallbackTenant()
```
mpanne, line 28: What are the cases when a `FallbackTenant` is returned? And what happens next? Doesn't the `SecurityHeaderFilter` throw an exception in that case since the `cspNonce` is `null`? Why not throwing an exception right away then?

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
```diff
@@ -20,26 +21,43 @@ import kotlin.text.Charsets.UTF_8
  * Documentation about matomo events
  * https://matomo.org/guide/reports/event-tracking/
  */
-@Profile("!local")
```
mpanne, line 23: Why removing this? We shouldn't try to send events to `example.org` locally imho.

src/main/kotlin/de/bund/digitalservice/useid/tracking/matomo/MatomoTrackingService.kt
mpanne: +1

mpanne (approved):
Some good cleanups. 
I don't fully understand the `ResolveTenantFilter` but everything else looks legit to me! Let's maybe talk about it. 
Thank you for your long breath with this topic!!

src/main/kotlin/de/bund/digitalservice/useid/metrics/MetricsService.kt
hannesschaletzky: Good idea. I have added this as a comment for improvement, since we want to merge first and then do more improvements.

src/main/kotlin/de/bund/digitalservice/useid/tracking/TrackingProperties.kt
hannesschaletzky: No, I removed that so we can send tracking requests locally to our useid-playground matomo site.

src/main/resources/application-local.yaml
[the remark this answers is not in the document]
pgurusinga: changing the structure of tenants can be challenging, but if you want to change this naming, we need to be very careful since you need to ensure the old code with old properties still works.

hannesschaletzky:
Last thing to do (fail fast improvement): 
We remove the FallbackTenant & WidgetDefaultTenant and we clarify in which cases we require a tenant and in which we don't. We don't set a tenant if none could be resolved. If a tenant is required and there is no tenant, throwing an exception is ok. 
We need to add the missing tests to verify the behaviour (Go through controller where tenants are required and test both cases[success, error]).

PR #214, opened 2023-02-28
USEID-712: Add CSP header to fallback page

mpanne:


mpanne:
Not the desired behaviour. We do not want to allow ANY page to include the fallback page in an iframe.

PR #222, opened 2023-03-16
USEID-869: Implement WebAuthn registration and authentication

mpanne:
# Secure device switch by implementing WebAuthn (Prototype)

## Goal
We believe that WebAuthn would bring a secure and user friendly solution to transfer information between mobile app and desktop app. Therefore, we would like to implement WebAuthn and try out, how it works and behaves.

## Description
Add package `credentials` which includes a new controller which handles requests from widget and eID-Client to register new credentials and authenticate for existing credentials.

## Note
This is a technical prototype which is hidden behind a feature flag. The code is not production ready and no tests exist.


src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -0,0 +1,87 @@
...
+
+internal const val CREDENTIALS_BASE_PATH = "/api/v1/credentials"
+
+@RestController
+@ConditionalOnProperty(name = ["features.desktop-solution-enabled"], havingValue = "true")
+class CredentialController(
+    private val credentialService: CredentialService,
+    private val eventService: EventService,
+) {
+    @PostMapping(path = [CREDENTIALS_BASE_PATH])
```
code28, line 25: Let's add correct swagger docs right away:
```
     @Operation(summary = "Start registration for webauthn credentials")
     @ApiResponse(responseCode = "202", content = [Content()])
```

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -0,0 +1,87 @@
...
+                    credential.credentialId,
+                    credential.pckCreationOptions.toCredentialsCreateJson(),
+                ),
+            )
+    }
+
+    @PutMapping(
+        path = ["$CREDENTIALS_BASE_PATH/{credentialId}"],
+        headers = ["Content-Type=application/json"],
+    )
```
code28, line 48: (no content of course only when the code is adjusted..)
```
     @ResponseStatus(HttpStatus.NO_CONTENT)
     @Operation(summary = "Complete webauthn registration")
     @ApiResponse(responseCode = "204", content = [Content()])
```

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -0,0 +1,87 @@
...
+            startRegistrationRequest.refreshAddress,
+        )
+
+        return ResponseEntity
+            .status(HttpStatus.ACCEPTED)
+            .contentType(MediaType.APPLICATION_JSON)
+            .body(
+                StartRegistrationResponse(
+                    credential.credentialId,
+                    credential.pckCreationOptions.toCredentialsCreateJson(),
```
code28, line 40:
```suggestion
                    credential.pkcCreationOptions.toCredentialsCreateJson(),
```

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialService.kt
```diff
@@ -0,0 +1,135 @@
...
+        credentialId: UUID,
+        publicKeyCredentialJson: String,
+    ): Credential {
+        val credential = credentialMockDatasource.findById(credentialId)
+            ?: throw CredentialNotFoundException(credentialId)
+
+        val pkc = PublicKeyCredential.parseRegistrationResponseJson(publicKeyCredentialJson)
+
+        val finishRegistrationOptions = FinishRegistrationOptions.builder()
+            .request(credential.pckCreationOptions)
```
code28, line 81:
```suggestion
            .request(credential.pkcCreationOptions)
```

src/main/resources/static/js/webauthn.js
```diff
@@ -0,0 +1,145 @@
+function registerWebAuthnCredentials(widgetSessionId) {
```
code28, line 1: Shall we put a divider commment between registration (app simulator) and authentication (widget)?

code28 (changes requested):
Really nice! 🚀
And very clean code, I like it!

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -0,0 +1,87 @@
...
+        @RequestBody publicKeyCredentialJson: String,
+    ): ResponseEntity<Any> {
+        credentialService.finishRegistration(credentialId, publicKeyCredentialJson)
+
+        val credential = credentialService.startAuthentication(credentialId)
+        sendAuthenticateEventToWidget(credential)
+
+        return ResponseEntity
+            .status(HttpStatus.ACCEPTED)
+            .contentType(MediaType.APPLICATION_JSON)
```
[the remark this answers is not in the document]
hannesschaletzky: +1

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -0,0 +1,87 @@
...
+            .build()
+    }
+
+    private fun sendAuthenticateEventToWidget(credential: Credential) {
+        eventService.publish(
+            AuthenticateEvent(credential.credentialId, credential.assertionRequest!!.toCredentialsGetJson()),
```
hannesschaletzky, line 66: I don't have a suggestion, but maybe we could get rid of the double bang operator.

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialService.kt
```diff
@@ -0,0 +1,135 @@
...
+    private fun buildUser(widgetSessionId: UUID): UserIdentity {
+        val username = widgetSessionId.toString()
+        return UserIdentity.builder()
+            .name(username)
+            .displayName(username)
+            .id(generateUserId())
+            .build()
+    }
+
+    private fun buildStartRegistrationOptions(user: UserIdentity): StartRegistrationOptions? {
```
hannesschaletzky, line 60: Why does this have an optional return type?

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialService.kt
```diff
@@ -0,0 +1,135 @@
...
+
+        return credentialMockDatasource.update(credential)
+    }
+
+    fun startAuthentication(credentialId: UUID): Credential {
+        val credential = credentialMockDatasource.findById(credentialId)
+            ?: throw CredentialNotFoundException(credentialId)
+
+        val assertionRequest: AssertionRequest = relyingParty.startAssertion(
+            StartAssertionOptions.builder().username(credential.username).build(),
```
hannesschaletzky, line 102: can we also do that builder pattern with a new line each time a new function is called? "Inline" builder patterns look weird to me

src/main/resources/static/js/webauthn.js
hannesschaletzky: +1

hannesschaletzky (approved):
Especially the CredentialController and the CredentialService reads like a charme 👍 very clean IMHO. 
I did not have a look onto the JS / app-simulator code.

src/main/resources/static/js/webauthn.js
hannesschaletzky: +1

src/main/kotlin/de/bund/digitalservice/useid/credentials/Credential.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.yubico.webauthn.data.PublicKeyCredentialDescriptor
+import java.util.UUID
+
+data class Credential(
+    val credentialId: UUID,
+    val widgetSessionId: UUID,
+    val username: String,
+    val userIdBase64: String,
+    val refreshAddress: String,
+    val pckCreationOptions: PublicKeyCredentialCreationOptions,
```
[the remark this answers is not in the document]
mpanne: thx, done

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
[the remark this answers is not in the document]
mpanne: Good catch, I cleaned that up.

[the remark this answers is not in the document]
mpanne: Meaning I added `produces` everywhere and removed `headers` which is set automatically returning a `ResponseEntity`.

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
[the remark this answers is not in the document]
mpanne: Yes, we should return 204 (see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT)):

> If the target resource does have a current representation and that representation is successfully modified in accordance with the state of the enclosed representation, then the origin server must send either a [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) (OK) or a [204](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) (No Content) response to indicate successful completion of the request.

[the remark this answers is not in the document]
mpanne: Done

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialMockDatasource.kt
```diff
@@ -0,0 +1,79 @@
...
+        return credentials.find { it.keyId?.id == keyCredentialId }
+    }
+
+    fun delete(credentialId: UUID) {
+        credentials.removeAll { it.credentialId == credentialId }
+    }
+
+    override fun getCredentialIdsForUsername(username: String): MutableSet<PublicKeyCredentialDescriptor> {
+        val credential = findByUsername(username) ?: return mutableSetOf()
+        return mutableSetOf(credential.keyId!!)
```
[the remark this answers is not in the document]
mpanne: What errors occurred when testing? Shouldn't the `keyId` always be present?

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
mpanne: Thanks, will do. Returning 202 makes no sense here, right? The entity is created right away, so we should return 201.

mpanne: Done

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialMockDatasource.kt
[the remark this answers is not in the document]
code28: I don't recall unfortunately. The `keyId` should be present actually - maybe it was a weird state, when testing. I'll see if I can reproduce it. But anyway, I think having as few force unwraps as possible is good, so I'd change that anyway. What do you think?

src/main/resources/static/js/webauthn.js
mpanne: I'm totally not a friend of divider comments. Reasons: 
- You always forget them when refactoring / moving code, so they lose their meaning and are more confusing than helpful pretty soon.
- There are functions in this file which are used by both, registration and authentication. Another separator for that?

mpanne: If you wanna separate it, I could create multiple files. 
(Totally out of scope: Once we have a build pipeline for our widget, we would split up those files anyways and have them re-combined during the build)

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
mpanne: Added a null check before, which throws a more descriptive exception.

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialService.kt
mpanne: Good catch. Auto extracted method. Removed.

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialService.kt
mpanne: Done

allowed-licenses.json
```diff
@@ -89,6 +89,13 @@
     },
     {
       "moduleLicense": "BSD Zero Clause License"
+    },
+    {
+      "moduleLicense": "BSD3"
+    },
+    {
+      "moduleLicense": null,
+      "moduleName": "com.yubico:webauthn-server-parent"
```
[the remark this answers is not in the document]
mpanne: Declares to use BSD license (see [here](https://mvnrepository.com/artifact/com.yubico/webauthn-server-parent/2.4.0)). BSD is compatible to EUPL in all versions according to [this matrix](https://joinup.ec.europa.eu/collection/eupl/matrix-eupl-compatible-open-source-licences).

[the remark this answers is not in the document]
mpanne: I added a note

code28 (approved):
I'd still vote for removing force unwrap if possible, even if I can't reproduce the error, but for me it's okay to keep it like that, until we can reproduce it. 🙃

src/main/resources/static/js/webauthn.js
mpanne: Got your point now. 😉 I will add comments to the methods only used for the app simulation.

PR #231, opened 2023-03-28
USEID-691: Migrate from Autent SDK to PANSTAR SDK

mpanne:
# Goal

Switch to new eID-Server PANSTAR from current Autent to be able to use the `TransactionInfo` as specified in the technical guidelines and make advantage of a more up-to-date software with less bugs.

# Description

* Include new PANSTAR SDK and remove Autent SDK
* Migrate code to use new SDK


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -42,21 +54,83 @@ class IdentificationSessionService(private val identificationSessionRepository:
         return "${applicationProperties.baseUrl}$IDENTIFICATION_SESSIONS_BASE_PATH/${session.useIdSessionId}/$TCTOKEN_PATH_SUFFIX"
     }
 
+    @Transactional
```
JulianRoesner, line 57: Does the whole function need to be transactional? We use that for the database, correct? I am wondering whether we could add this to a function that wraps the parts that are executing the database calls?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -42,21 +54,83 @@ class IdentificationSessionService(private val identificationSessionRepository:
...
-        val identificationSession = identificationSessionRepository.findByUseIdSessionId(useIdSessionId)
-            ?: throw IdentificationSessionNotFoundException(useIdSessionId)
-        val tcToken: TCTokenType
+        val requestData: RequestData = getRequestData(useIdSessionId)
+
+        val tcTokenWrapper: TcTokenWrapper?
         try {
-            val eidService = EidService(eidServiceConfig, identificationSession.requestDataGroups)
-            tcToken = eidService.getTcToken("${applicationProperties.baseUrl}$REFRESH_PATH")
+            tcTokenWrapper = soapHandler.getTcToken(requestData, "${applicationProperties.baseUrl}$REFRESH_PATH")
```
JulianRoesner, line 63: I am wondering whether we could rename the soapHandler to something more specific: eIdSoapHandler maybe?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -42,21 +54,83 @@ class IdentificationSessionService(private val identificationSessionRepository:
...
         } catch (e: Exception) {
             tcTokenCallsWithErrorsCounter.increment()
             log.error("Failed to get tc token for identification session. useIdSessionId=$useIdSessionId", e)
             throw e
         }
-        updateEIdSessionId(useIdSessionId, extractEIdSessionId(tcToken))
-        return tcToken
+
+        val session = identificationSessionRepository.findByUseIdSessionId(useIdSessionId)
+        session!!.eIdSessionId = extractEIdSessionId(tcTokenWrapper.tcToken)
```
JulianRoesner, line 72: The actual update is happening through the transactional, is that correct? Why do we want to get rid of the update function? Because this is now the only place where it is needed? 
Personally, I think having the function call helps a lot with readability. wdyt?

JulianRoesner (commented):
Looks good as far as I understand it. I had some questions. If you want to we can also quickly chat to answer those. Whatever works good for you.

When we will actually merge this, we should probably add an ADR. This could potentially also include note on the way we chose to load the certificates etc..

As this is still a draft, I assume that we will probably have another round of rework before merging this?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -116,40 +99,23 @@ class IdentificationSessionsController(
...
             .status(HttpStatus.OK)
             .contentType(MediaType.APPLICATION_JSON)
             .body(userData)
     }
+
+    fun validateApiKey(authentication: Authentication, eIdSessionId: UUID) {
+        val apiKeyDetails = authentication.details as ApiKeyDetails
+
+        val identificationSession = identificationSessionService.findByEIdSessionId(eIdSessionId)
+            ?: throw IdentificationSessionNotFoundException()
```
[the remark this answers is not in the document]
mpanne: Correct!

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -116,40 +99,23 @@ class IdentificationSessionsController(
...
-        if (userData.result.resultMajor.equals("http://www.bsi.bund.de/ecard/api/1.1/resultmajor#ok")) {
-            try {
-                identificationSessionService.delete(identificationSession)
-            } catch (e: Exception) {
-                log.error("Failed to delete identification session. id=${identificationSession.id}")
-            }
-        } else {
-            // resultMinor error codes can be found in TR 03130 Part 1 -> 3.4.1 Error Codes
-            log.info("The resultMinor for identification session is ${userData.result.resultMinor}. id=${identificationSession.id}")
-        }
```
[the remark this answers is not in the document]
mpanne: Moved to `IdentificationSessionService`

JulianRoesner:
> Very good comments @JulianRoesner. I stopped reading through it at some point since this is only a draft and was implemented quick and dirty to test the migration to the PANSTAR. Those comments will be valuable once we plan to bring this to production.


Sounds good 👍

PR #238, opened 2023-04-05
NO-TICKET: Finish refactoring of identification package

mpanne:


src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -80,8 +85,46 @@ class IdentificationSessionService(
         return session
     }
 
-    fun delete(identificationSession: IdentificationSession) {
-        identificationSessionRepository.delete(identificationSession)
-        log.info("Deleted identification session. useIdSessionId=${identificationSession.useIdSessionId}")
+    @Transactional
```
JulianRoesner, line 88: I was wondering the same for the PANSTART draft. Maybe now is the better point to ask:
- Why do we need to set transactional here?
- Why do we set it for the whole function and not only the part that actually does sth with the database?
   -  Is that because we need to keep getEidInformationa and deleteSession together transactional?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -116,38 +102,20 @@ class IdentificationSessionsController(
...
         return ResponseEntity
             .status(HttpStatus.OK)
             .contentType(MediaType.APPLICATION_JSON)
-            .body(userData)
+            .body(identity)
+    }
+
+    fun validateTenant(tenant: Tenant, eIdSessionId: UUID) {
+        val identificationSession = identificationSessionService.findByEIdSessionId(eIdSessionId)
+            ?: throw IdentificationSessionNotFoundException()
```
JulianRoesner, line 116: Is there any place where we would not throw the IdentificationSessionNotFounException after calling findByEIdSessionId? If so, we could maybe encapsulate that part?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -116,38 +102,20 @@ class IdentificationSessionsController(
...
-            try {
-                identificationSessionService.delete(identificationSession)
-            } catch (e: Exception) {
-                log.error("Failed to delete identification session. id=${identificationSession.id}")
-            }
-        } else {
-            // resultMinor error codes can be found in TR 03130 Part 1 -> 3.4.1 Error Codes
-            log.info("The resultMinor for identification session is ${userData.result.resultMinor}. id=${identificationSession.id}")
-        }
+        val identity = identificationSessionService.getIdentity(eIdSessionId, tenant.id)
```
JulianRoesner, line 107: 👍

JulianRoesner (commented):
LGTM! I am only wondering about the use of the transactional

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
```diff
@@ -38,8 +36,7 @@ class RefreshController(private val identificationSessionService: Identification
         @RequestParam requestQueryParams: Map<String, String>,
     ): ResponseEntity<Unit> {
         val session = identificationSessionService.findByEIdSessionId(eIdSessionId) ?: run {
-            log.error("Failed to load identification session.")
-            return ResponseEntity.status(HttpStatus.NOT_FOUND).build()
+            throw IdentificationSessionNotFoundException()
```
code28, line 39: We don't need that `run` thing and can just use `?: throw ...`

code28 (approved):
This looks so good and clean. Love it.

src/main/kotlin/de/bund/digitalservice/useid/refresh/RefreshController.kt
mpanne: Yes. Resolved by moving this to the `findByEIdSessionId` method (see comment above).

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionRepository.kt
```diff
@@ -10,5 +11,9 @@ interface IdentificationSessionRepository : JpaRepository<IdentificationSession,
     fun findByEIdSessionId(eIdSessionId: UUID): IdentificationSession?
     fun findByUseIdSessionId(useIdSessionId: UUID): IdentificationSession?
     fun deleteAllByCreatedAtBefore(before: LocalDateTime)
+
+    @Modifying
+    @Query("DELETE FROM IdentificationSession session WHERE session.eIdSessionId = :eIdSessionId")
+    fun deleteByEIdSessionId(eIdSessionId: UUID)
```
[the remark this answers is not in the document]
mpanne: It's not the same thing. This part here removes identification sessions from the database which are completed successfully and therefore not needed anymore. The schedular removes idle session which have been started but never completed. 
I would still vote for deleting the completed once immediately since there is no reason to keep them.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
mpanne: I read again [this documentation](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#transactions) and would therefore propose to set the transaction method next to the `@Modifying` query on the repository method. The transaction is needed here since we are using a custom `@Query` and  Spring does not know ootb if it's a `readOnly` method anymore.

JulianRoesner: Thanks. That makes sense to me 👍

PR #240, opened 2023-04-05
USEID-712: Add CSP header for fallback page

code28:


src/main/kotlin/de/bund/digitalservice/useid/config/WidgetSecurityHeadersFilter.kt
```diff
@@ -2,36 +2,37 @@ package de.bund.digitalservice.useid.config
...
         }
 
-        if (!pathIsValidWidgetPages) return filterChain.doFilter(request, response)
-
-        val tenant = request.getAttribute(REQUEST_ATTR_TENANT) as Tenant
-        tenant.cspNonce = UUID.randomUUID().toString()
-        response.setHeader(HTTP_HEADER_CONTENT_SECURITY_POLICY, WidgetContentSecurityPolicy.headerValue(tenant.cspHost, tenant.cspNonce))
-        response.setHeader(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, tenant.cspHost)
-        response.setHeader(HttpHeaders.VARY, HttpHeaders.ORIGIN)
+        if (requestPathMatches(FALLBACK_PAGE, request)) {
```
JulianRoesner, line 28: Is there a situation where we would have both (WIDGET_PAGE and FALLBACK_PAGE)? Otherwise, we could use `else if` here. WDYT?

hannesschaletzky: +1

src/main/kotlin/de/bund/digitalservice/useid/config/WidgetSecurityHeadersFilter.kt
```diff
@@ -2,36 +2,37 @@ package de.bund.digitalservice.useid.config
...
+            response.setHeader(HTTP_HEADER_CONTENT_SECURITY_POLICY, WidgetContentSecurityPolicy.headerValueNoFrame())
```
hannesschaletzky, line 29: is there a reason why there are no tests for that in WidgetControllerIntegrationTest?

hannesschaletzky, line 29: We probably need the nonce logic here as well, since we execute inline js. I posted the error in slack in the thread.   
If you run the app locally with the e-service-example, and you go onto the fallback page, you will see the error in your console.

hannesschaletzky (changes requested):
I think we need the nonce logic here as well because of the inline-js ☺️

src/main/kotlin/de/bund/digitalservice/useid/config/WidgetSecurityHeadersFilter.kt
hannesschaletzky: image

src/main/kotlin/de/bund/digitalservice/useid/config/WidgetSecurityHeadersFilter.kt
code28: Nope, just thought, it is independent from each other, but actually a switch would be the best (not possible here I think), so else-if comes closest to switch 👍

src/main/kotlin/de/bund/digitalservice/useid/config/WidgetSecurityHeadersFilter.kt
code28: Haha, I was looking everywhere for tests for this file but couldn't find any. 😆 Thanks for pointing me to that, of course I want to add a test! :)

code28: You're right. Thought it had something to do with the widget-script, but it's our inline script, right.

PR #241, opened 2023-04-06
NO-TICKET: Dependencies update

pgurusinga:
Because of https://github.com/digitalservicebund/useid-backend-service/pull/236, they need to be updated together (after building the backend locally)


pgurusinga:
already updated

PR #248, opened 2023-04-19
USEID-920: Migrate TC-Token endpoint

UrsKahmann:
Related PR: https://github.com/digitalservicebund/useid-eservice-example/pull/22

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
@@ -26,7 +26,6 @@ class SecurityConfig(
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
         return http.authorizeHttpRequests()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tc-token").permitAll()
             .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens/*").permitAll()
             .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens").permitAll()
             .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_OLD_BASE_PATH/*/transaction-info").permitAll()
--- a/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
@@ -1,7 +1,6 @@
 package de.bund.digitalservice.useid.config
 
 import de.bund.digitalservice.useid.identification.IDENTIFICATIONS_BASE_PATH
-import de.bund.digitalservice.useid.identification.IDENTIFICATIONS_OLD_BASE_PATH
 import de.bund.digitalservice.useid.tenant.MANAGE_IDENTIFICATION_SESSION_AUTHORITY
 import de.bund.digitalservice.useid.tenant.ResolveTenantFilter
 import de.bund.digitalservice.useid.tenant.TenantAuthenticationFilter
@@ -26,13 +25,11 @@ class SecurityConfig(
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
         return http.authorizeHttpRequests()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens/*").permitAll()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens").permitAll()
-            .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_OLD_BASE_PATH/*/transaction-info").permitAll()
+            .requestMatchers("$IDENTIFICATIONS_BASE_PATH/*/tokens/*").permitAll()
+            .requestMatchers("$IDENTIFICATIONS_BASE_PATH/*/tokens").permitAll()
+            .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_BASE_PATH/*/transaction-info").permitAll()
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").authenticated()
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/**").authenticated()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/**").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
             .anyRequest().permitAll()
             .and().csrf().disable()
             .headers()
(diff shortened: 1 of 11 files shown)


src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -25,7 +26,7 @@ class SecurityConfig(
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
         return http.authorizeHttpRequests()
-            .requestMatchers("$IDENTIFICATION_SESSIONS_BASE_PATH/*/tc-token").permitAll()
+            .requestMatchers("$TCTOKEN_BASE_PATH/*").permitAll()
```
code28, line 29: Can be removed (hopefully, please validate).

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -19,20 +19,14 @@ import org.springframework.security.core.Authentication
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
-import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import java.util.UUID
 
 internal const val IDENTIFICATION_SESSIONS_BASE_PATH = "/api/v1/identification/sessions"
-internal const val TCTOKEN_PATH_SUFFIX = "tc-token"
+internal const val TCTOKEN_BASE_PATH = "/api/v1/tc-tokens"
```
code28, line 26: I'd also prefer plural here ("TCTOKENS")

code28 (approved):
One thing as discussed, one proposed renaming, otherwise 👍

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
mpanne: That's right. 
This made me thinking if we shouldn't require authentication by default and just explicitly set `permitAll` on public endpoints?

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
```diff
@@ -52,6 +49,10 @@ class IdentificationSessionsController(private val identificationSessionService:
         content = [Content()],
     )
     @SecurityRequirement(name = "apiKey")
+    @Tag(
+        name = "Identification Sessions",
+        description = "An identification session represent an ongoing identification flow of a user and stores the required information.",
+    )
```
mpanne, line 55: I feel like this should stay on class level. I would vote for a separate `TCTokenController` to separate concerns here.

mpanne: Plus, if we wanna be consistent and use the new grouping by actor (see tag name "eID-Client" below) for our API docs, this tag name would need to be "eService".

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
mpanne: But out of scope for this PR probably.

mpanne (changes requested):
Congrats to you first PR @UrsKahmann! Good work. I would suggest some refactoring though.

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
code28: Good point, I like that from a security perspective. Since we only have very few endpoints with authentication it might be less code for a list like this. But we should take a look at this - after migrating all the endpoints and not in the scope of this PR, I'd say.

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionController.kt
code28: I agree, that having this on class level feels better, but thinking about e.g. the time-based tokens, we have two endpoints and each one for a different actor. So we have to decide: Grouping by context internally and by actor in the docs OR grouping by actor everywhere OR grouping by actor in the docs and otherwise a mix of both 😅. I'd actually vote for the first option and think it's not too confusing to do it like that. What do you think?

Still, for this specific case, I think it's valid to have its own TCTokenController separately! 👍

mpanne: Agree on everything.

JulianRoesner (commented):
LGTM. I like the refactoring (especially in the tests). This makes it a lot more readable imo.

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -26,14 +25,11 @@ class SecurityConfig(
...
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tc-token").permitAll()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens/*").permitAll()
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/*/tokens").permitAll()
-            .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_OLD_BASE_PATH/*/transaction-info").permitAll()
+            .requestMatchers("$IDENTIFICATIONS_BASE_PATH/*/tokens/*").permitAll()
+            .requestMatchers("$IDENTIFICATIONS_BASE_PATH/*/tokens").permitAll()
+            .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_BASE_PATH/*/transaction-info").permitAll()
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").authenticated()
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
-            .requestMatchers("$IDENTIFICATIONS_OLD_BASE_PATH/**").authenticated()
```
code28, line 35: We still need this, as long as the old endpoints are still available. Or am I missing something?! 🤔

src/main/kotlin/de/bund/digitalservice/useid/identification/TcTokenController.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+internal const val TCTOKENS_BASE_PATH = "/api/v1/tc-tokens"
+
+@RestController
+@Timed
+@Tag(name = "eID-Client")
```
code28, line 20: It's only one endpoint, but still, adding `RequestMapping` including the base path might make sense.

code28 (changes requested):
Please revisit the security config or tell me, what I am missing :)

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
UrsKahmann: @JulianRoesner

code28: I just saw, that meanwhile the old endpoints are removed.. so nevermind

code28 (approved):
Adding `RequestMapping` might make sense, but I leave it up to you to decide if you want to add it.
I also really like the refactoring in the tests! 👍

PR #251, opened 2023-04-21
USEID-923: Migrate transactions endpoints

UrsKahmann:
Migrated the transaction endpoints to the new [scheme](https://digitalservicebund.atlassian.net/wiki/spaces/UseID/pages/628392825/API+Design)

What else needs to be done: 
- [x] Adapt endpoint in eService Example app ([done here](https://github.com/digitalservicebund/useid-eservice-example/pull/23))
- [x] Adapt path in iOS prototype


src/main/kotlin/de/bund/digitalservice/useid/transactioninfo/TransactionInfoController.kt
```diff
@@ -12,18 +13,22 @@ import org.springframework.web.bind.annotation.RequestBody
...
 
 @RestController
-@Tag(name = "Transaction Info", description = "Additional information regarding the identification which will be displayed in the eID-Client.")
 @ConditionalOnProperty(name = ["features.desktop-solution-enabled"], havingValue = "true")
 class TransactionInfoController(
     private val transactionInfoService: TransactionInfoService,
 ) {
     @PostMapping(
-        path = ["/api/v1/identification/sessions/{useIdSessionId}/$TRANSACTION_INFO_SUFFIX"],
+        path = ["/api/v1/identifications/{useIdSessionId}/$TRANSACTION_INFO_SUFFIX"],
```
JulianRoesner, line 24: I am wondering whether we should move "/api/v1" into it's own constant 🤔 wdyt?

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -25,9 +26,10 @@ class SecurityConfig(
     @Bean
     fun securityFilterChain(http: HttpSecurity): SecurityFilterChain {
         return http.authorizeHttpRequests()
-            .requestMatchers(HttpMethod.GET, "$IDENTIFICATIONS_BASE_PATH/*/transaction-info").permitAll()
```
JulianRoesner, line 28: We can remove this, because we are permitting all requests anyway, correct?

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
```diff
@@ -25,9 +26,10 @@ class SecurityConfig(
...
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").authenticated()
             .requestMatchers("$IDENTIFICATIONS_BASE_PATH/**").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
+            .requestMatchers(HttpMethod.POST, "/api/v1/identifications/*/$TRANSACTION_INFO_SUFFIX").authenticated()
+            .requestMatchers(HttpMethod.POST, "/api/v1/identifications/*/$TRANSACTION_INFO_SUFFIX").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
```
JulianRoesner, line 32: Why do we have to add this? I would have assumed that the pattern in line 29 and 30 would match already also this endpoint. Otherwise, we would change behavior as the only the name of the endpoints has changed. Or am I missing something?

If we keep it, could we use the constant:
```suggestion
            .requestMatchers(HttpMethod.POST, "$IDENTIFICATIONS_BASE_PATH/*/$TRANSACTION_INFO_SUFFIX").authenticated()
            .requestMatchers(HttpMethod.POST, "$IDENTIFICATIONS_BASE_PATH/*/$TRANSACTION_INFO_SUFFIX").hasAuthority(MANAGE_IDENTIFICATION_SESSION_AUTHORITY)
```

JulianRoesner (commented):
LGTM so far. I have only some questions and I am not so sure about the security config

src/main/kotlin/de/bund/digitalservice/useid/transactioninfo/TransactionInfoController.kt
```diff
@@ -12,18 +13,22 @@ import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
 import java.util.UUID
 
-internal const val TRANSACTION_INFO_SUFFIX = "transaction-info"
+internal const val TRANSACTION_INFO_SUFFIX = "transaction-infos"
```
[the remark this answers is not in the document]
JulianRoesner: I've just renamed it.

src/main/kotlin/de/bund/digitalservice/useid/config/ApplicationProperties.kt
```diff
@@ -11,4 +11,8 @@ import org.springframework.validation.annotation.Validated
 class ApplicationProperties {
     @NotBlank
     lateinit var baseUrl: String
+
+    companion object {
+        const val apiVersionPrefix = "/api/v1"
+    }
```
pgurusinga, line 17: my lack of knowledge: why must it be a companion object?

is it possible to do like this?
image

JulianRoesner: That's the part I am also unsure about. I started with sth. similar to your approach. Unfortunately, this did not work because we use it assign it a lot to `const val` and we need to assign a constant value to that. We need `const val` because we set it a lot in the Controller annotations where we need constants at compile time. 
 However, we can only add a constant value at the top-level. So, the other way instead of having a companion object that I see would be to add it as a top level constant value in the Application Properties. 
That said, I am very new to Kotlin. Therefore, I have probably missed sth and I am not sure what would be the best way to stay true to Kotlin.

I love the variable name btw ;)

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
UrsKahmann: Yes, at least that's my understanding of it.

src/main/kotlin/de/bund/digitalservice/useid/config/ApplicationProperties.kt
pgurusinga: thanks for the explanation!

src/main/kotlin/de/bund/digitalservice/useid/config/SecurityConfig.kt
JulianRoesner: Talked to @mpanne  and we can remove those lines

PR #252, opened 2023-04-25
USEID-1025: Change tags for CredentialController

JulianRoesner:
I've added new tags and some documentation for the webauthn prototype routes. Does this make sense to you?

### DIFF
--- a/src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
+++ b/src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
@@ -24,10 +24,6 @@ internal const val CREDENTIALS_BASE_PATH = "/api/v1/credentials"
 @RestController
 @RequestMapping(CREDENTIALS_BASE_PATH)
 @ConditionalOnProperty(name = ["features.desktop-solution-enabled"], havingValue = "true")
-@Tag(
-    name = "Credentials",
-    description = "WebAuthn credentials are used to authenticate the a user in the widget.",
-)
 class CredentialController(
     private val credentialService: CredentialService,
     private val eventStreamService: EventStreamService,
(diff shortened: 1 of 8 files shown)


src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
```diff
@@ -24,17 +24,17 @@ internal const val CREDENTIALS_BASE_PATH = "/api/v1/credentials"
...
 class CredentialController(
     private val credentialService: CredentialService,
     private val eventStreamService: EventStreamService,
 ) {
     @PostMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
     @Operation(summary = "Start registration for WebAuthn credentials.")
     @ApiResponse(responseCode = "201", content = [Content()])
+    @Tag(
+        name = "eID-Client",
+        description = "WebAuthn credentials are created in the app to later authenticate the user in the widget.",
```
mpanne, line 36: this unfortunately hasn't the desired effect. Endpoints are grouped by tag. The tag can only have one description. The first description detected is used. In this case you would override the description of the `eID-Client` tag (the whole group of endpoints), not the description of the method (which is in line 32). 
Screenshot 2023-04-27 at 15 19 52

mpanne: So please remove all descriptions here.

mpanne (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/credentials/CredentialController.kt
JulianRoesner: I moved the descriptions for those and the other tags into custom annotations. This should make it more clear that those are only used once and where to update the descriptions.
Bildschirmfoto 2023-04-28 um 13 54 20

PR #254, opened 2023-04-26
USEID-835: Change port of local database to non-default

JulianRoesner:
Am I missing something? Imo the testing container does not need to be changed because the port is set automatically.


mpanne (approved):
LGTM

hannesschaletzky (approved):
LGTM2

PR #259, opened 2023-05-05
Useid-811: Add eid healtcheck endpoint

JulianRoesner:
This adds an additional endpoint to check whether the eId-Server is available. This is available under `api/v1/eidservice/health`.  It will check in the redis database whether the last entries to the eIdServer failed. If that value is below a configurable threshold, it responds with `DOWN`. Otherwise, it responds with `UP`.

The eIdServer is tested every minute and the result is stored in redis. The entries in redis have a ttl of 5 minutes. We wanted to make that configurable. However, I could not find a good way to do it as we configure the TTL through an annotation which needs to be compile constant and therefore can not be an env value.

We could use this endpoint in uptime to monitor the health of our application (s. [this pr](https://github.com/digitalservicebund/uptime-monitor/pull/214)).

# Feedback
- Do you think the naming of the endpoint makes sense?
- Where could the naming in general be improved?
- Is it fine that the considered timeframe is not configurable/do you have an idea to make it easily configurable?


JulianRoesner (commented):
@mpanne and I did a pair review yesterday to check this PR.

PR #272, opened 2023-05-19
USEID-1045: Fix missing tenantId in prometheus metrics endpoint

hannesschaletzky:
[Ticket](https://digitalservicebund.atlassian.net/jira/software/projects/USEID/boards/3?label=Backend&selectedIssue=USEID-1045)

Goal: 
As a product manager or engineer, I want to be able to differentiate metrics by tenants to see which tenant has the highest response times. 

What did we do?
Urs and I replaced the deprecated CustomWebClientExchangeTagsProvider.kt with a new RequestMetrictsTenantTagProvider.kt and adjusted the unit tests and added an integration test. 

Some routes are difficult to track. For the TCTokenURL we decided that we attach the tenantId as a query parameter. The URL now looks as follows 
`"http://localhost:8080/api/v1/tc-tokens/e5d5c59f-6966-40ed-ae6e-593ec32dd9de?tenant_id=tenantFoo"`
WDYT? 

Result: 
tenantId appears in our metrics for all endpoints where a tenant object is found in the request object. Attached a screenshot for the create session endpoint with multiple tenants. 
image


hannesschaletzky:
main had too many differences due to test folder restructuring + hamcrest and jupiert.assertions removal. 
created [new PR ](https://github.com/digitalservicebund/useid-backend-service/pull/273)

PR #273, opened 2023-05-19
USEID-1045: Fix missing tenantId in prometheus metrics endpoint

hannesschaletzky:
[Ticket](https://digitalservicebund.atlassian.net/jira/software/projects/USEID/boards/3?label=Backend&selectedIssue=USEID-1045)

Goal: 
As a product manager or engineer, I want to be able to differentiate metrics by tenants to see which tenant has the highest response times. 

What did we do?
Urs and I replaced the deprecated CustomWebClientExchangeTagsProvider.kt with a new RequestMetrictsTenantTagProvider.kt and adjusted the unit tests and added an integration test. 

Some routes are difficult to track. For the TCTokenURL we decided that we attach the tenantId as a query parameter. The URL now looks as follows 
`"http://localhost:8080/api/v1/tc-tokens/e5d5c59f-6966-40ed-ae6e-593ec32dd9de?tenant_id=tenantFoo"`
WDYT? 

Result: 
tenantId appears in our metrics for all endpoints where a tenant object is found in the request object. Attached a screenshot for the create session endpoint with multiple tenants. 
image


hannesschaletzky:
We still have a failing test:
`✘ prometheus log contains server request metric with tenant ID for widget endpoint after widget is requested()`

hannesschaletzky:
During the build, the tests pass. 
```
    ✔ prometheus log contains server request metric with tenant ID for tc token endpoint after tc token is requested()
    ✔ prometheus log contains server request metric with tenant ID for start session endpoint after start session call()
    ✔ prometheus log contains server request metric with tenant ID for widget endpoint after widget is requested()
```

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
```diff
@@ -41,7 +42,7 @@ class IdentificationSessionService(
     }
 
     private fun buildTcTokenUrl(session: IdentificationSession): String {
-        return "${applicationProperties.baseUrl}$TCTOKENS_BASE_PATH/${session.useIdSessionId}"
+        return "${applicationProperties.baseUrl}$TCTOKENS_BASE_PATH/${session.useIdSessionId}?$PARAM_NAME_TENANT_ID=${session.tenantId}"
```
mpanne, line 45: By appending the `tenant_id` to the tc token URL we are externalizing a value which is meant for internal purposes and tracking. I think we should find another solution to track the tenant for the tc token request. The tenant could for example be derived from the `useIdSessionId` which is included in the request. The `ResolveTenantFilter` could do that for example. Please reconsider this.

mpanne (changes requested):

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
hannesschaletzky: Urs and I thought about this as well. By selecting the tenant via `useIdSessionId` within the `RequestMetricsTenantTagProvider`, we need to inject the repository and make another database call. 
However, maybe we find a way to attach the `tenantId` to the request within the `TcTokenController`. 
Prometheus uses two calls to track a route. A beginning call (start time), then logic, then ending call (stop time). Maybe we can read the attached `tenantId` at the ending call. Will try out.

src/integrationTest/kotlin/de/bund/digitalservice/useid/tenant/RequestMetricsTenantIdTagIntegrationTest.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    @Test
+    fun `prometheus metric contains tenant ID tag for tc token endpoint`() {
+        // Given
+        val expectedPrometheusLogRegex = Regex("http_server_requests_seconds_count\\{error=\"none\",exception=\"none\",method=\"GET\",outcome=\"SUCCESS\",status=\"200\",tenant_id=\"(.+)\",uri=\"/api/v1/tc-tokens/\\{useIdSessionId}\",}")
+
+        // When
+        var tcTokenURL = ""
+        webTestClient.sendStartSessionRequest()
+            .expectBody().jsonPath("$.tcTokenUrl").value<String> { tcTokenURL = it }
```
kaibernhard, line 59: This request could be "given":
_Given_ a session has been started
_when_ I get a tc token
_then_ the prometheus log contains the tenant id

src/integrationTest/kotlin/de/bund/digitalservice/useid/identification/TestHelper.kt
```diff
@@ -18,7 +18,7 @@ internal fun WebTestClient.sendStartSessionRequest() =
         .headers { setAuthorizationHeader(it) }
         .exchange()
 
-internal fun WebTestClient.sendGETRequest(uri: String) =
+internal fun WebTestClient.createGETRequest(uri: String) =
```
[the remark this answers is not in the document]
mpanne: Cool, I agree!

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
mpanne: This is better. Could you maybe extract that into a method which says what it does so it's better understandable later?
And isn't that too late to be included in the tag for the metric? Isn't the metric already sent? 

Some more information about an additional database call: 
[Hibernate (ORM used by Spring data) caches all session objects in a "first level cache"](https://www.digitalocean.com/community/tutorials/hibernate-caching-first-level-cache). So requesting the identification session twice during one requests only results in one database query since afterwards the entity is in the cache. Hence, performance-wise this does not make a difference. So this might be something worth considering again?

UrsKahmann:
> We still have a failing test in step "analyse" in the pipeline. `✘ prometheus log contains server request metric with tenant ID for widget endpoint after widget is requested()`
> 
> I can't figure out why. Locally it works fine.

I will investigate this further 👍

src/main/kotlin/de/bund/digitalservice/useid/identification/IdentificationSessionService.kt
hannesschaletzky: Thanks for sharing. Nice first level caching! Did not know. 
However, it feels a bit odd passing the repository through the app. It would also require us to add more logic in the tags provider. So I guess attaching the tenantId from the controller to the request seems like the best solution. 
I don't fully understand what you mean by "it is too late", since the tenantId is being attached to the metrics, thus, it works. 
Will extract the logic into a function with Urs, though :)

UrsKahmann:
> > We still have a failing test in step "analyse" in the pipeline. `✘ prometheus log contains server request metric with tenant ID for widget endpoint after widget is requested()`
> > I can't figure out why. Locally it works fine.
> 
> I will investigate this further 👍

Okay so I am actually clueless why this fails. AS @hannesschaletzky also tested, it works locally. And when running the pipeline multiple times it occasionally goes through successfully.. Seems like it is flaky for some reason but I have no clue how to investigate further 🤔

hannesschaletzky:
We will debug together tomorrow @UrsKahmann.

hannesschaletzky:
Urs and I need help on this one... 

We are trying to print the debug logs but it actually now fails during the build with a spotless failure. However, running spotless locally does not change anything and the spotless checks within our commit-hook are successful as well. 
Only within the pipeline it fails. We can't figure out why.

PR #278, opened 2023-05-23
NO-TICKET: Use GitHub Gradle Action

kaibernhard:
See https://github.com/gradle/gradle-build-action#why-use-the-gradle-build-action

Setting the executable permission for gradlew has been removed since the permission is set correctly by Git.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -50,10 +50,9 @@ jobs:
         with:
           java-version: "17.0"
           distribution: "temurin"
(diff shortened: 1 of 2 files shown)


kaibernhard:
> I like! Thanks for cleaning this up. Do you see improvements in terms of performance due to the new cache? Or any other noticable advantages so far?

In this job, the build was actually 1 min slower than one random other PR, but I assume the cache was completely cold for this run. We don't have enough data to answer the question yet.

PR #294, opened 2023-06-06
Switch to MIT

ChaosCoder:
We can switch to MIT as our dependencies use weak copyleft (EUPL, LGPL, GPL+Classpath Exception) or more permissive licenses (MIT, Apache, BSD, CC0).


JulianRoesner (approved):
👍

